PR #1, opened 2016-04-26
Init

ImageLabel:
contribution :+1:


SinisterRectus:
The upRank and downRank methods should be for ServerChannel, not TextChannel, as you can change both text and voice channel positions. Notice that I already implemented channel editing shortly before your commit. See https://github.com/SinisterRectus/Discordia/commit/2b7d521e0c68e3b41865bdbb835d06a80bc774a6. I did want to include moveUp and moveDown methods, but I did not for now, because I was having trouble implementing it properly. Will revisit in the future.

I can add your examples, but I might forget to maintain them while working on the library itself.

ImageLabel:
Ooh - didn't think much about the up and down methods. Also didn't see your earlier commit so i'll just close this request. As for the example, I haven't tested it but can't see any blatant error off the bat so it's up to you whether to add it or not. Still looking forward to contributing to this project for practice and  knowledge if you don't mind.

SinisterRectus:
No worries. Since, at this stage, changes are likely to happen frequently, it's probably best to check with me before starting work on a contribution.

PR #4, opened 2016-05-03
Adding nickname support for Client and User

XNBlank:


SinisterRectus:
Two things:
- The method name should be setNickname() for clarity.
- The method should be in the Member class, not User. Nicknames can only be set for members of a server.

XNBlank:
whoops my bad. Lemme fix it quick.

XNBlank:
Alright fixed I believe.

PR #6, opened 2016-05-19
Usermanagement

sclark39:
Adding partial permission support to discordia and adding some user management stuff (create roles, assign roles to members)

also adds support for Member:setMute, Member:setDeaf, Role:setPermissions, Member:set { options }, and Role:set {options}

### DIFF
--- a/discordia/classes/discord/member.lua
+++ b/discordia/classes/discord/member.lua
@@ -28,9 +28,23 @@ function Member:_update(data)
 	self.gameName = data.game and data.game.name or self.gameName-- string or nil
 end
 
-function Member:setNickname(nickname)
-	local body = {nick = nickname or ''}
-	self.client:request('PATCH', {endpoints.servers, self.server.id, 'members', self.id}, body)
+-- Member:set* Functions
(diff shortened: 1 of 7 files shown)


SinisterRectus:
- `shallowCopyArray` is redundant. see `table.copy` in `extensions.lua`.
- I don't do any type checking anywhere, so there's no need to implement it now. If you do want it, though, I have a `Error` and `Warning` classes that do not terminate the program. There is currently no difference between the two except for the text coloring and implied severity.
- Formatting
  - Inconsistencies with library: `( param )` vs `(param)` and `i,v` vs `i, v`.
  - Don't use snake_case if you can avoid it (ex: `target_type`)
  - Use UpperCase for class names
  - Use lowercase for file names
  - Use camelCase for everything else

sclark39:
Cool, I'll fix all that.

sclark39:
Let me know if there is anything else that needs adjusting. I am happy to do so.

XNBlank:
Damn, beat me to it. :P

PR #10, opened 2016-06-21
Add docs to some files

LoganDark2:
I've added some docs to some files, as that's easier than adding it to the wiki.

I'll continue to add more docs as I get time to.


LoganDark2:
Actually, I'm going to develop offline until the docs are complete.

LoganDark2:
edit: sinisterrectus doesn't like docs in his code

SinisterRectus:
That's not true. I said that if we're going to have comment docs, then they should autogenerate external docs (via ldoc).

PR #12, opened 2016-06-23
Add new example

LoganDark2:
This shows how message.acknowledge should be used and is also a helpful little selfbot as well. Win win!


LoganDark2:
Example is extremely unsecure (password stored)
Will update and then reopen

PR #16, opened 2016-10-10
2 Extra self data.

PurgePJ:


SinisterRectus:
`User.id` and `User.discriminator` are already defined. They do not require accessor attributes.

PurgePJ:
Right.

PR #17, opened 2016-10-11
Added more table extensions. (1 improved)

PurgePJ:


SinisterRectus:
I already have a `table.copy`, and it is purposefully distinct from `table.deepcopy`. If you meant to replace the original, you did not remove it from the code.

I don't like how your `table.contains` can check nested tables one layer deep.

`table.clear` is okay, but not enough to accept the PR.

It's a good idea to contact me before submitting features in PRs.

PR #18, opened 2016-10-11
Create Invite.lua

PurgePJ:
Hopefully everything is correct.

### DIFF
--- a/discordia/containers/snowflakes/Invite.lua
+++ b/discordia/containers/snowflakes/Invite.lua
@@ -1,0 +1,35 @@
+local Snowflake = require('../Snowflake')
+
+local Invite = class('Invite', Snowflake)
+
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Copy-pasted pre-1.0 code is not going to work with 1.0. There have been changes to the Discord API (removal of xkcd codes, inclusion of vanity URLs) since this was written. Also, data is handled differently in 1.0.

Invites do not have Snowflake IDs, so I would like for them to not inherit from the Snowflake class if possible.

PurgePJ:
It's not actually fully copy pasted, added there accept and decline functions and also adapted two vars to 1.0but well I didn't remove the not needed data. Mistake.

PurgePJ:
Instead of decline -> delete

PR #20, opened 2016-11-02
Update reply.lua

999gary:
adding :new() call or else it returns an error


Tigerism:
Are you on v1.00? The :new() is only used on pre-v1.

SinisterRectus:
`Client:new()` is pre-1.0 behavior. Use `Client()` in 1.0 and later.

PR #21, opened 2016-11-04
Updated the example to 1.0 + added one more event

PurgePJ:


SinisterRectus:
Pulled for now, will probably edit it in the future. Thanks.

PR #24, opened 2016-11-23
Update extensions.lua

nolivesmatter:
+adds table.slice


SinisterRectus:
I'm not going to pull code that is copied from the first StackOverflow result via Google Search. Check the coding style of Discordia's extensions and adjust your code so that it matches.

SinisterRectus:
Implemented as of commit d04fd9d.

PR #25, opened 2016-11-23
update extension.lua

nolivesmatter:
adds startsWith >.>

`function string.startsWith(str, pat)
	return ((str:sub(1, pat:len()) ==pat and true) or false)
end`

### DIFF
--- a/discordia/extensions.lua
+++ b/discordia/extensions.lua
@@ -135,6 +135,10 @@ end
 
 -- string --
 
+function string.startsWith(str, pat)
+	return ((str:sub(1, pat:len()) ==pat and true) or false)
+end
+
 function string.split(str, delim)
 	if delim and delim ~= '' then
 		local words = {}


SinisterRectus:
This can be done trivially using `string.find(str, pattern) == 1` or `str:find(pattern) == 1`

nolivesmatter:
it's all for the feel y'know, and it's more friendly to people who are more custom to node.js

SinisterRectus:
I understand the sentiment. Many Luvit features are designed to be similar to Node features, but `string.startsWith` is not a Node feature; it is a feature of JavaScript. Lua is not JavaScript, and `string.find` is simple enough. Maybe I'll add `startswith` at a later date, but I would not add it without `endswith`, which I don't find very useful. If you can come up with a reasonable use for both, then I may add them. (Remember, Lua is about simplicity, and I've already added more bloat to Discordia than I'd like at this point).

SinisterRectus:
I went against my original judgement and added `string.startswith` and `string.endswith` as of commit d04fd9d. Note the lowercase, which matches Lua's standard library.

PR #26, opened 2016-11-24
Send Embed support

PurgePJ:
Approving this file change proposal you will:
1- let us send embed messages
2- once you update Discordia some users, including me, will not have to update this file again to be able to send embeds as we already are doing.

I'm sure you'll find a better way to add embed message but this is a workaround until that moment. Almost every lib is supporting sending embeds.

### DIFF
--- a/discordia/containers/snowflakes/channels/TextChannel.lua
+++ b/discordia/containers/snowflakes/channels/TextChannel.lua
@@ -73,7 +73,7 @@ local function getPinnedMessages(self)
 	return _messageIterator(self, client._api:getPinnedMessages(self._id))
 end
 
-local function sendMessage(self, content, mentions, tts)
+local function sendMessage(self, content, embed, mentions, tts)
 	if type(mentions) == 'table' then
 		local strings = {}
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Embeds will be added, but I cannot accept this PR for the following reasons:

- As written, this proposed change breaks backwards compatibility, and does not include documentation. The embed parameter will either have to go at the end of the sendMessage, or we will have to include a second function such as sendMessageEmbed(), sendEmbedMessage(), sendEmbeddedMessage(), or whatever else sounds good.
- Incoming embeds have existed for a while, but I still need to assess whether I want to write an embed object wrapper. Working with raw tables or arrays is not idiomatic for Discordia, and I try to avoid it whenever possible.
- Outgoing embed payloads have only been official for two days, and I am still on a break for the holiday.

If you are bugged by updates that lack this feature, and if you want to send a raw payload, then you may want to (in your bot code) call the API method directly, until I figure out how to implement a cleaner method,
```lua
client._api:createMessage(channel.id, {
  content = content, embed = embed, -- etc
})
```

PR #27, opened 2016-12-01
Small change in API Header

PurgePJ:
Added the possibility to swap from Content-Type depending on the function. 

It's not required to add the last arg but for sending files you do need to change Content-Type header to multipart/form-data.


SinisterRectus:
Since Discordia does not yet handle file uploads, we don't need this yet. If I wait to pull it until after we implement file uploading, there will be merge conflicts. Please contact me before making further additions. It took me longer to read these commits than it would have taken to write the code that they provide.

PR #30, opened 2016-12-27
Webhook addition.

PurgePJ:
The following **files** contain what is necessary in order to implement Webhook functions for Discord.


SinisterRectus:
Sorry it's taken so long to review this. The code style looks great on the surface, but there is too much nonsensical code here, and I don't have the time to point out all the issues.

PR #31, opened 2016-12-27
Files in order to support Embed.

PurgePJ:


discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
...
+local function getGuild(self)
+	return self._parent._api:getGuildEmbed(self._guild_id)
+end
+
+local function modifyGuild(self, enabled, channel_id)
+	return self._parent._api:modifyGuildEmbed(self._guild_id, {enabled = enabled, channel_id = channel_id})
+end
+
+property('enabled', '_enabled', nil, 'string', "Wether the Embed is enabled or not.")
+property('channel_id', '_channel_id', nil, 'string', "Embed's channel id (if enabled is true).")
```
SinisterRectus: Expose as channel, not ID. See parent chain for other objects.

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
...
+	if not self._enabled then return end 
+	self._channel_id = data.channel_id
+end
+
+function Embed:__tostring()
+	return format('%s: %s', self.__name, self._channel_id or "") -- will not always have a channel id
+end
+
+
+local function getGuild(self)
```
SinisterRectus: Why does this use the API? Use the cached Guild object.

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
...
+Embed.__description = "Represents Discord Embed."
+
+function Embed:__init(data, parent)
+	Snowflake.__init(self, data, parent)
+	self._enabled = data.enabled
+	if not self._enabled then return end 
+	self._channel_id = data.channel_id
+end
+
+function Embed:__tostring()
```
SinisterRectus: `__tostring` methods should not return an empty string. See examples in other locations.

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
+local Snowflake = require('./Snowflake')
+
+local format = string.format
+
+local Embed, property, method = class('Embed', Snowflake)
+Embed.__description = "Represents Discord Embed."
+
+function Embed:__init(data, parent)
```
SinisterRectus: Why are you manually copying values? This is done automatically when `Container` is constructed.

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
...
+	Snowflake.__init(self, data, parent)
+	self._enabled = data.enabled
+	if not self._enabled then return end
```
SinisterRectus: Logic is not necessary here.

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
+local Snowflake = require('./Snowflake')
+
+local format = string.format
+
+local Embed, property, method = class('Embed', Snowflake)
+Embed.__description = "Represents Discord Embed."
```
SinisterRectus: Change to "Represents a Discord message embed."

discordia/containers/Embed.lua
```diff
@@ -0,0 +1,35 @@
...
+function Embed:__tostring()
+	return format('%s: %s', self.__name, self._channel_id or "") -- will not always have a channel id
+end
+
+
+local function getGuild(self)
+	return self._parent._api:getGuildEmbed(self._guild_id)
+end
+
+local function modifyGuild(self, enabled, channel_id)
```
SinisterRectus: Why is this named "modifyGuild"? It modifies the embed. Regardless, Discordia doesn't use "modify" methods. See `Guild` for an example of how all the properties that can be modified are split into different methods.

discordia/containers/snowflakes/Guild.lua
```diff
@@ -386,6 +386,36 @@ local function findMembers(self, predicate)
 	return self._members:findAll(predicate)
 end
 
+-- embed --
+
+local function getEmbed(self)
```
SinisterRectus: Does this return more than one object? Why does it have a singular name?

discordia/containers/snowflakes/Guild.lua
```diff
@@ -386,6 +386,36 @@ local function findMembers(self, predicate)
...
+	return function()
+		local v = data[i]
+		if v then
+			i = i + 1
+			return Embed(v, client)
+		end
+	end
+end
+
+local function modifyEmbed(self, payload)
```
SinisterRectus: What is this and why is it not `Embed:modify`?

discordia/containers/snowflakes/channels/TextChannel.lua
```diff
@@ -101,6 +101,15 @@ local function sendMessage(self, content, mentions, tts)
 	if success then return self._messages:new(data) end
 end
 
+local function sendEmbed(self, embed, content)
```
SinisterRectus: I need to rework message sending. Embed sending will be integrated into existing `sendMessage` methods.

discordia/containers/snowflakes/Guild.lua
PurgePJ: It returns one object; it has got a singular name since you only pass One Embed Object to this class.
Of course a message can have more than one Embed but in this case you are not passing the whole message object and checking then for Embeds.

discordia/containers/Embed.lua
PurgePJ: It is called "modifyGuild" because what you are modifying is that guild's Embed. Said that, I understand what you mean by splitting into different methods and I will add it right now.

SinisterRectus:
Note: This is for "guild embeds". These are distinct from message "embeds", which are the subject of your original `sendEmbed` method and current developments. We may include guild embed objects at a later date, as indicated in `API.lua`.

SinisterRectus:
I don't even know what to do with this one. We don't need this feature now, or in the immediate future, anyway.

PR #33, opened 2016-12-28
File uploading implemented.

PurgePJ:


discordia/client/Multipart.lua
```diff
@@ -0,0 +1,16 @@
+local Multipart = {}
```
SinisterRectus: There's no reason to have a separate file and table for one function. It can be integrated into `TextChannel` as a local function.

discordia/containers/snowflakes/channels/TextChannel.lua
```diff
@@ -101,6 +102,12 @@ local function sendMessage(self, content, mentions, tts)
 	if success then return self._messages:new(data) end
 end
 
+local function sendFile(self, data, name, field)
```
SinisterRectus: Remove this. We will integrate it into `sendMessage`.

discordia/client/API.lua
```diff
@@ -569,6 +574,11 @@ end
 
 -- end of auto-generated methods --
 
+function API:uploadFile(channel_id, payload)
```
SinisterRectus: Remove this. We will integrate it into `createMessage`.

discordia/client/Multipart.lua
PurgePJ: I was following other lib's structure, the same way I got inspired on doing this.

discordia/client/API.lua
```diff
@@ -72,7 +72,12 @@ function API:request(method, route, endpoint, payload)
 	end
 
 	if method:find('P') then
-		payload = payload and encode(payload) or '{}'
+		if isMulti then
+			reqHeaders[3][2] = 'multipart/form-data; boundary=Discordia'
```
SinisterRectus: This is not good. There is no guarantee that the third header is always `content-type`.

PurgePJ: In case it's a file uploading, yes it will be.

SinisterRectus: It's not for two reasons. The iterator order of `pairs` cannot be trusted and self._headers` may not always been the same. The appropriate header needs to be explicitly located. You can't assume that it's always the third one.

SinisterRectus:
Requires higher-level abstraction. Will implement in due time.

PR #41, opened 2017-03-01
Changed string.split extension to work with patterns

Siapran:


discordia/extensions.lua
```diff
@@ -1,6 +1,6 @@
 local random = math.random
 local insert, remove, sort, concat = table.insert, table.remove, table.sort, table.concat
-local gmatch, match, byte, char = string.gmatch, string.match, string.byte, string.char
+local gmatch, match, byte, char, find, sub = string.gmatch, string.match, string.byte, string.char, string.find, string.sub
```
SinisterRectus: `find` is defined in the row below this

discordia/extensions.lua
```diff
@@ -137,9 +137,20 @@ end
 
 function string.split(str, delim)
 	if delim and delim ~= '' then
+		if find(str, delim) == nil then
```
SinisterRectus: Change to `if not find(str, delim) then return {str} end` to match styles

discordia/extensions.lua
```diff
@@ -137,9 +137,20 @@ end
...
-		for word in gmatch(str .. delim, '(.-)' .. delim) do
+		local pattern = '(.-)' .. delim
+		local lastPos = 1
+		while true do
+			local _, pos, word = find(str, pattern, lastPos)
+			if not word then break end
 			insert(words, word)
+			lastPos = pos + 1
+		end
+		if lastPos <= #str then
```
SinisterRectus: This is not necessary if you append `delim` to `str` before the loop, as in the original function.

Siapran: Try the original with "%s+" as the delimiter:
```string.split("arg1 \narg2	 arg3", "%s+")```

File: discordia/extensions.lua (hunk 2 above)
Comment (Siapran) [inline discordia/extensions.lua]: Fixed

File: discordia/extensions.lua (hunk 3 above)
Comment (SinisterRectus) [inline discordia/extensions.lua]: I didn't test it with a pattern.

Comment (Siapran) [inline discordia/extensions.lua]: The point of this change is to allow the string.split extension to work with patterns.

Comment (SinisterRectus) [inline discordia/extensions.lua]: Obviously. I was just stating a fact.

**Comment by SinisterRectus** [comment]
I want to say to just use `for word in str:gmatch("([^" .. delim .. "]+)") do` since it's a lot simpler and probably more efficient. It allows for pattern splitting, and it prevents empty strings from being matched. This may break backwards compatibility for people who are expecting empty strings, though.

**Comment by SinisterRectus** [comment]
There are lots of different ways to split strings. Maybe a string.split2 can be added.

**Comment by Siapran** [comment]
Maybe make string.split work with plain delimiters by default, and add a `pattern` parameter to enable pattern splitting?

**Comment by SinisterRectus** [comment]
For something like an extension module, I'd prefer if it were added as a separate function.

**Comment by SinisterRectus** [comment]
I think should use @FiniteReality's version. It's more efficient than all others so far.

```lua
if delim and delim ~= '' then
    local current = 1
    local result = { }
    while true do
        local start, finish = find(str, delim, current)
        if start and finish then
            insert(result, sub(str, current, start-1))
            current = finish + 1
        else
            break
        end
    end
    if current <= #str then
        insert(result, sub(str, current))
    end
    return result
else
	local chars = {}
	for char in gmatch(str, '.') do
		insert(chars, char)
	end
	return chars
end
```

SinisterRectus:
Change your function name to `string.split2` and I'll pull it, then work on changing it to the updated version.

Siapran:
@FiniteReality's version is faster and produces the same output as my version. So yes, definitely use it.

PR #53, opened 2017-05-27
Added 'streaming' status

ProtectiveManEgg:
I meant to do this a while ago for you, but I forgot about it, and just finally got around to cutting out the emoji stuff.

Could probably be cleaned up and made a lot better, but this is mostly just for a base line, and maybe done properly in Discordia 2.0?


SinisterRectus:
I would have accepted this a month ago, but not now. I do not want to release any more minor versions of Discordia between 1.5 and 2.0. If I do not include this in 2.0, then you can PR it there when it is made available.

PR #65, opened 2017-08-28
Fix for the 'August 16, 2017 — Breaking Change'

PurgePJ:
Minimal but enough changes were made to fix setting a game name to the client.


SinisterRectus:
We're too close to a 2.0 release to bother with maintaining what will soon become legacy code. The master branch will receive no updates and 1.5.1 will be the last version before 2.0.

PR #70, opened 2017-09-11
Added an early exit and modified returning message for example in README.md

antonio6643:
To prevent it from continuously calling itself when the command is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ client:on('messageCreate', function(message)
 	if message.author == client.user then return end
 	
 	if message.content == '!ping' then
-		message.channel:sendMessage('Pong!')
+		message.channel:sendMessage('!pong')
 	end
 	
 end)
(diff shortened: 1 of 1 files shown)


Tigerism:
The reply, ``!pong``, is not the same as the command, ``!ping``

antonio6643:
Oh, I thought it was just a regular response. My mistake, I edited the file to put it back as it was.

SinisterRectus:
The readme example is meant to be as simple as possible while still being functional, so I purposefully excluded the author check. This is found, and explained, in the examples folder instead.

I would rather see the change from "!pong" to "pong!", though.

antonio6643:
That makes sense. I'll edit the request again to add it back.

SinisterRectus:
You missed the point of my comment. I don't want the early exit in this example.

antonio6643:
My mistak. My power is out so I'm doing this all from mobile.

SinisterRectus:
Remove the new lines.

PR #90, opened 2017-11-30
Added missing WebhookClient

PurgePJ:


libs/client/API.lua
```diff
@@ -657,37 +657,37 @@ function API:getWebhookWithToken(webhook_id, webhook_token) -- not exposed, need
 	return self:request("GET", endpoint)
 end
 
-function API:modifyWebhook(webhook_id, payload) -- Webhook:_modify
+function API:modifyWebhook(webhook_id, payload) -- WebhookClient:modify
```
SinisterRectus: Please undo this change.

libs/client/API.lua
```diff
@@ -657,37 +657,37 @@ function API:getWebhookWithToken(webhook_id, webhook_token) -- not exposed, need
...
 end
 
-function API:modifyWebhookWithToken(webhook_id, webhook_token, payload) -- not exposed, needs webhook client
+function API:modifyWebhookWithToken(webhook_id, webhook_token, payload) -- WebhookClient:modifyWebhookWithToken
 	local endpoint = f(endpoints.WEBHOOK_TOKEN, webhook_id, webhook_token)
 	return self:request("PATCH", endpoint, payload)
 end
 
-function API:deleteWebhook(webhook_id) -- Webhook:delete
+function API:deleteWebhook(webhook_id) -- WebhookClient:delete
```
SinisterRectus: Please undo this change.

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+function get.name(self)
+	return self._options.name
+end
+
+function get.token(self)
+	return self._options.token
+end
+
+-- Setters and Modifiers
+function WebhookClient:modify(tbl)
```
SinisterRectus: This should be a private method `_modify`

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+	local data, err = self._api:modifyWebhookWithToken(self._options.id, self._options.token, tbl)
+	if data then 
+		return true
+	else
+		return false, err
+	end
+end
+
+function WebhookClient:setAvatar(avatar)
+	avatar = Resolver.base64(avatar) or resolveImage(avatar) or null
```
SinisterRectus: Why is `resolveImage` used here? `avatar` should be base64 or an image path.

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+
+function get.token(self)
+	return self._options.token
+end
+
+-- Setters and Modifiers
+function WebhookClient:modify(tbl) 
+	assert(type(tbl) == "table", "Invalid modify parameter. Table expected")
+
+	for k, v in pairs(tbl) do
```
SinisterRectus: It's not necessary to check for channel_id, just ignore it like I do with all of the other PATCHes.

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+	return self._options.name
+end
+
+function get.token(self)
+	return self._options.token
+end
+
+-- Setters and Modifiers
+function WebhookClient:modify(tbl) 
+	assert(type(tbl) == "table", "Invalid modify parameter. Table expected")
```
SinisterRectus: This check is not necessary.

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+
+	local data, err = self:modify({avatar = avatar})
+	if data then
+		return true
+	else
+		return nil, err
+	end
+end
+
+function WebhookClient:setChannelId(channelId)
```
SinisterRectus: I thought you can't change the channel?

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+	}, webhookData.file, wait)
+
+	if data then
+		return true
+	else
+		return nil, err
+	end
+end
+
+function WebhookClient:delete()
```
SinisterRectus: Not necessary to have two methods. Just use `delete`. Kind of a bad name, though. It looks like you're deleting the Client.

libs/client/WebhookClient.lua
```diff
@@ -0,0 +1,274 @@
...
+	return self:deleteWithToken()
+end
+
+function WebhookClient:deleteWithToken()
```
SinisterRectus: See above.

libs/client/WebhookClient.lua
PurgePJ: If you were going to delete the client I would use ``destroy``. Delete is used to delete the webhook from discord.

SinisterRectus: Maybe `deleteWebhook` to indicate that you are actually deleting the webhook and not the client. I do not like "destroy".

SinisterRectus: But then `setName` and `setAvatar` become confusing because you're setting the properties of the webhook, not the client. Hmmm.

PurgePJ: In the end we need to have in mind that we are managing a webhook using an imitation of a Client. I will change ``delete`` to ``deleteWebhook``

SinisterRectus:
I will address this after voice is mostly completed.

trumpCoder:
I guess this should be added... Other libraries have this.

trumpCoder:
Please add this in. Today

Issue #78, opened 2017-10-07
Too many members


SinisterRectus:
If a member leaves a guild, its object is uncached normally. After this, a presence update can come in that signifies that the member's status has changed to offline. If the client has `cacheAllMembers` enabled, the presence update will add a new member to the guild. The guild's cache will then have more members than it should. This becomes a critical issue when a member leaves a guild while member chunks are still being received. Unfortunately, there is no straightforward way to distinguish between members that have left the guild, members that are actually going offline, and members that are switching to invisible.

jo0sh:
Oh

jo0sh:
When will this be fixed

SinisterRectus:
As soon as I find the time to fix it.

SinisterRectus:
Issue still exists. Will have to re-write presence handling some time soon.

trumpCoder:
Uhh, then if I get this "too many members" error, does that mean this library doesn't support large servers?

trumpCoder:
This must be fixed ASAP.

SinisterRectus:
The issue should be fixed as of b286af2f2d34988402d4664692e08a37a6e79527. I'm not sure when I'll be able to publish a release to lit, but you're free to use the GitHub version instead.

SinisterRectus:
Released in version 2.3.0.

PR #91, opened 2017-12-01
Fix for Issue #78

truemedian:
Quick fix that should handle Issue #78 in most cases.

### DIFF
--- a/libs/client/EventHandler.lua
+++ b/libs/client/EventHandler.lua
@@ -269,7 +269,7 @@ function EventHandler.GUILD_EMOJIS_UPDATE(d, client)
 	return client:emit('emojisUpdate', guild)
 end
 
-local presence_fix = setmetatable({}, {__index = function(self, k) self[k] = {}; return self[k] end)
+local presence_fix = setmetatable({}, {__index = function(self, k) self[k] = {}; return self[k] end})
 function EventHandler.GUILD_MEMBER_ADD(d, client)
 	local guild = client._guilds:get(d.guild_id)
 	if not guild then return warning(client, 'Guild', d.guild_id, 'GUILD_MEMBER_ADD') end
--- a/libs/client/EventHandler.lua
+++ b/libs/client/EventHandler.lua
@@ -269,11 +269,15 @@ function EventHandler.GUILD_EMOJIS_UPDATE(d, client)
 	return client:emit('emojisUpdate', guild)
 end
 
+local presence_fix = setmetatable({}, {__index = function(self, k) self[k] = {}; return self[k] end)
 function EventHandler.GUILD_MEMBER_ADD(d, client)
 	local guild = client._guilds:get(d.guild_id)
 	if not guild then return warning(client, 'Guild', d.guild_id, 'GUILD_MEMBER_ADD') end
 	local member = guild._members:_insert(d)
 	guild._member_count = guild._member_count + 1
+	if client._options.cacheAllMembers then
+		presence_fix[guild.id][d.user.id] = nil
+	end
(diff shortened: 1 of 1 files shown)


libs/client/EventHandler.lua
```diff
@@ -284,11 +284,16 @@ function EventHandler.GUILD_MEMBER_UPDATE(d, client)
...
 
+local presence_fix = {}
 function EventHandler.GUILD_MEMBER_REMOVE(d, client)
 	local guild = client._guilds:get(d.guild_id)
 	if not guild then return warning(client, 'Guild', d.guild_id, 'GUILD_MEMBER_REMOVE') end
 	local member = guild._members:_remove(d)
 	guild._member_count = guild._member_count - 1
+	if client._options.cacheAllMembers then
+		presence_fix[guild.id] = presence_fix[guild.id] or {}
+		presence_fix[guild.id][d.user.id] = true
```
SinisterRectus: For consistency, this should be nil'd on GUILD_MEMBER_ADD.

libs/client/EventHandler.lua
```diff
@@ -426,6 +431,14 @@ function EventHandler.PRESENCE_UPDATE(d, client) -- may have incomplete data
 		else
 			member = guild._members:get(d.user.id)
 			if not member then
+				if client._options.cacheAllMembers then
+					presence_fix[guild.id] = presence_fix[guild.id] or {}
+					if presence_fix[guild.id][d.user.id] then
+						presence_fix[guild.id][d.user.id] = nil
+						client:debug('Ignoring PRESENCE_UPDATE for user \'%s\' in guild \'%s\'', d.user.id, guild.id)
+						return
```
SinisterRectus: Move return up for a proper tail-call.

libs/client/EventHandler.lua
```diff
@@ -284,11 +284,16 @@ function EventHandler.GUILD_MEMBER_UPDATE(d, client)
...
 end
 
+local presence_fix = {}
 function EventHandler.GUILD_MEMBER_REMOVE(d, client)
 	local guild = client._guilds:get(d.guild_id)
 	if not guild then return warning(client, 'Guild', d.guild_id, 'GUILD_MEMBER_REMOVE') end
 	local member = guild._members:_remove(d)
 	guild._member_count = guild._member_count - 1
+	if client._options.cacheAllMembers then
+		presence_fix[guild.id] = presence_fix[guild.id] or {}
```
SinisterRectus: Might want to use an auto-expanding table. See listenersMeta in Emitter.lua.

libs/client/EventHandler.lua
```diff
@@ -426,6 +431,14 @@ function EventHandler.PRESENCE_UPDATE(d, client) -- may have incomplete data
 		else
 			member = guild._members:get(d.user.id)
 			if not member then
+				if client._options.cacheAllMembers then
+					presence_fix[guild.id] = presence_fix[guild.id] or {}
```
SinisterRectus: Might want to use an auto-expanding table. See listenersMeta in Emitter.lua.

truemedian:
was a quick fix, will move to integrate it more to the lib

SinisterRectus:
Were you able to test this to see if it fixes the issue?

truemedian:
Yes, after receiving PRESENCE_UPDATE it does ignore the event for that member

SinisterRectus:
Does the member count stay accurate?

truemedian:
Hmm, good question, will test

truemedian:
After some testing on multiple guilds, I haven't had it fail; its giving accurate member counts
should work in all cases

PR #93, opened 2017-12-29
Adding JSON support to all Discordia Objects

truemedian:
Discordia Objects, such as Guild may be passed through json.stringify() without error, returning a valid JSON string

### DIFF
--- a/libs/iterables/ArrayIterable.lua
+++ b/libs/iterables/ArrayIterable.lua
@@ -22,7 +22,7 @@ function ArrayIterable:__len()
 	end
 end
 
-function ArrayIterable:__serializeJSON(null, cahce)
+function ArrayIterable:__serializeJSON(null)
 	local objects = {}
 	if self._map then
 		return wrap(function()


SinisterRectus:
I didn't realize you were manually including all of the properties. That can be done programmatically, and it probably should be. I'm not interested in maintaining the JSON objects for every Discordia object.

Why `__json` instead of `__tojson`?

Why do you return Lua tables? As far as I know, the methods should return strings so that they can be appended to and concatenated from a JSON buffer.

What is your use for this? Why should `Guild:__tojson` return all properties instead of just the ID, for example?

truemedian:
I was looking for a way to get all objects into json, am working on programmatically making json references atm

SinisterRectus:
I will address this after voice is mostly completed.

trumpCoder:
This sounds cool to have

SinisterRectus:
Here's the deal with serialization of Discordia objects. Let's take a color as an example:

```lua
local color = discordia.Color.fromHex('#FF8800')

for k, v in pairs(color) do
  print(k, v)
end
```

Currently, if you try to iterate over an obj with `pairs`, you will get the raw data stored in the object, with numerical indices. This is almost completely useless, since neither user nor library code should ever use these indices directly:
```lua
1       16746496
```

An abstraction level above this would expose the names of the raw data that point to the numerically indexed data. This is the first level of abstraction that the library uses, but users are told to not use these properties. Being able to see what is stored is nice for debugging purposes, but not for human-reading:
```lua
'_value'    16746496
```

Another level above this would iterate over all properties defined by Discordia class "getters". This is the highest level of abstraction relevant to a user inspecting an object. It provides all user-accessible properties; even the ones that have nil values, which is a rare treat for a Lua table. This is what I think we should expose:
```lua
'r'     255
'b'     0
'g'     136
'value' 16746496
```

These can all be done easily by modifying only `class.lua`.

----

If we do choose to provide a method for serializing class objects, we need to decide how the data should be presented. Should the method just return an iterator? Should it return a new table that holds the values? Should nested objects be returned as raw objects, should they be serialized, or should they just be tostring'd?

I'm not too interested in adding `__tojson` methods that are consumable by the built-in dkjson module. I don't want to make assumptions about what the user wants vs what Discordia wants. For example, the library would want `User:__tojson()` to return only the snowflake id, while the user may expect something different, like just the name, or perhaps all of the user's properties.

trumpCoder:
I need this for my bot. I want to see this added today.

PR #101, opened 2018-01-10
Fix for tokens

trumpCoder:
Token support is bugged. This pull request fixes https://github.com/SinisterRectus/Discordia/issues/100

You're welcome.

PR #104, opened 2018-02-07
Fix support for tokens

trumpCoder:
Bot tokens are bugged on the current discordia vers. This PR fixes that for you.

Your welcome.


SinisterRectus:
I already denied this in #101. This is not a bug.

PR #110, opened 2018-03-21
License update for 2018

Sanjay-B:

PR #112, opened 2018-03-21
Updated Package.lua to include 2018

Sanjay-B:


SinisterRectus:
Duplicate of #110 and #111

PR #120, opened 2018-04-30
Removed unnecessary check.

PurgePJ:
Removed unnecessary check. Following function style like lines 37 and 42.


SinisterRectus:
The check prevents the creation and use of an empty table.

PR #121, opened 2018-05-02
Improve message mention parsing

Mishio595:
Pattern was tested using
```lua
local function match(str, pat)
    return p(str, str:match(pat))
end

local function test(pat)
    print(pat)
    match('emoji <:name:1234>', pat)
    match('anim  <a:name:1234>', pat)
    match('user  <@1234>', pat)
    match('nick  <@!1234>', pat)
    match('role  <@&1234>', pat)
    match('chan  <#1234>', pat)

    print('Angle Bracket')
    match('< emoji <:name:1234>', pat)
    match('< anim  <a:name:1234>', pat)
    match('< user  <@1234>', pat)
    match('< nick  <@!1234>', pat)
    match('< role  <@&1234>', pat)
    match('< chan  <#1234>', pat)

    print('Punct')
    match('<- emoji <:name:1234>', pat)
    match('<- anim  <a:name:1234>', pat)
    match('<- user  <@1234>', pat)
    match('<- nick  <@!1234>', pat)
    match('<- role  <@&1234>', pat)
    match('<- chan  <#1234>', pat)
    print('')
end

test("<%P?(%p+)%w-%p?(%d+)>")
```

And it yields the output
```
<%P?(%p+)%w-%p?(%d+)>
'emoji <:name:1234>'    ':'     '1234'
'anim  <a:name:1234>'   ':'     '1234'
'user  <@1234>' '@'     '1234'
'nick  <@!1234>'        '@!'    '1234'
'role  <@&1234>'        '@&'    '1234'
'chan  <#1234>' '#'     '1234'
Angle Bracket
'< emoji <:name:1234>'  ':'     '1234'
'< anim  <a:name:1234>' ':'     '1234'
'< user  <@1234>'       '@'     '1234'
'< nick  <@!1234>'      '@!'    '1234'
'< role  <@&1234>'      '@&'    '1234'
'< chan  <#1234>'       '#'     '1234'
Punct
'<- emoji <:name:1234>' ':'     '1234'
'<- anim  <a:name:1234>'        ':'     '1234'
'<- user  <@1234>'      '@'     '1234'
'<- nick  <@!1234>'     '@!'    '1234'
'<- role  <@&1234>'     '@&'    '1234'
'<- chan  <#1234>'      '#'     '1234'
```

### DIFF
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -36,7 +36,7 @@ local function parseMentions(content)
 	local users, roles, channels, emojis = {}, {}, {}, {}
 	local seen = {}
 
-	for symbol, id in content:gmatch('<%P?(%p+)[^:]-%p?(%d+)>') do
+	for symbol, id in content:gmatch('<%P?(%p+)[%w_]-%p?(%d+)>') do
 		if not seen[id] then
 			if symbol == '@&' then
 				insert(roles, id)
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -36,7 +36,7 @@ local function parseMentions(content)
 	local users, roles, channels, emojis = {}, {}, {}, {}
 	local seen = {}
 
-	for symbol, id in content:gmatch('<.-(%p+)(%d+)>') do
+	for symbol, id in content:gmatch('<%P?(%p+)%w-%p?(%d+)>') do
 		if not seen[id] then
 			if symbol == '@&' then
 				insert(roles, id)
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -36,7 +36,7 @@ local function parseMentions(content)
 	local users, roles, channels, emojis = {}, {}, {}, {}
 	local seen = {}
 
-	for symbol, id in content:gmatch('<%P?(%p+)%w-%p?(%d+)>') do
+	for symbol, id in content:gmatch('<%P?(%p+)[^:]-%p?(%d+)>') do
 		if not seen[id] then
 			if symbol == '@&' then
 				insert(roles, id)


SinisterRectus:
The new pattern fails the Punct tests.

SinisterRectus:
```lua
'chan  <#1234>' '#'     '1234'
'chan  <<#1234>'        '<#'    '1234'
```
This breaks it. I'm thinking that there is no one pattern that can properly capture all mentions. I'll have to think about this for a bit.

SinisterRectus:
Try `<a?:?[%w_]-([^%P<]+)(%d+)>`

SinisterRectus:
This would probably be the best to use:
```lua
  for mention in content:gmatch('%b<>') do
    local symbol, id = mention:gmatch('<.-(%p+)(%d+)>')
```
From https://www.lua.org/manual/5.1/manual.html#5.4.1
>%bxy, where x and y are two distinct characters; such item matches strings that start with x, end with y, and where the x and y are balanced. This means that, if one reads the string from left to right, counting +1 for an x and -1 for a y, the ending y is the first y where the count reaches 0. For instance, the item %b() matches expressions with balanced parentheses.

SinisterRectus:
Eh, `'%b<>'` doesn't work as planned with nested mentions such as `<<#1234>>`. If you can get it to work somehow, great, otherwise, use a fancy capture.

SinisterRectus:
If you can show that the LPEG version performs significantly better (faster) than Lua's gmatch, then we might use it.

Mehgugs:
Initially the gmatch was ~x2 faster than my implementation but i removed the positional captures and it competes only being some ms slower, you can see here, if you can create a pattern which is able to cope with :

```
The mention string inserted randomly into 100000 bytes of lorem ipsum was: "<> <<a:#1234>> <<#1234><<@1234>><@!1234><@&1234<#1234<:name:1234>>><a:name:12344>>>"
This contains 5 correct mentions
buffer_size: 669125
Performing 500 tests of each method
Finished collecting data
Avg lpeg speed over 500 tests: 28.858531ms
Avg gmatch speed over 500 tests: 22.267634ms
lpeg passed the capture test
gmatch failed the capture test
```

Mehgugs:
After some further optimisation and tweaking:
```
The mention string inserted randomly into 100000 bytes of lorem ipsum was: "<@!word:1234><a@!word1234><><<a:#1234>><##1234><a::name:1234><<#1234><<@1234>><@!1234><@&1234<#1234<:name:1234>>><a:name:12344>>>"
This contains 5 correct mentions
Overall there are 34590 mentions within the buffer
buffer_size: 992748
Performing 1000 tests of each method
Finished collecting data
Avg lpeg speed over 1000 tests: 18.817483ms
Avg gmatch speed over 1000 tests: 35.535822ms
Avg rex speed over 1000 tests: 35.437953ms
```
[new parser](https://hastebin.com/moperigoqe.lua)

I also changed the parser to return a mentions object which has id arrays (or `{name, id}` pairs for -mojis) in `mentions[mention_type]`, this means it is similar in functionality to the current parseMentions function. 

```lua
  --example output
 parser([[<> <<a:#1234>> <<#1234><<@1234>><@!1234><@&1234<#1234<:name:1234>>><a:name:12344>>><@&1234>]])
----
{ animoji = { { 'name', '12344' } }, channels = { '1234' }, users = { '1234', '1234' }, roles = { '1234' }, emoji = { { 'name', '1234' } } }
```

SinisterRectus:
Can you run a benchmark on 1000 iterations of the original `parseMentions` function from `Message.lua` using your fancy content string?

SinisterRectus:
You might also want to switch your IDs to more unique ones since the "seen" check is omitting the repeats. This would not happen in practice, since channels, users, and regular roles will never have matching IDs.

SinisterRectus:
```lua
local function generateContent(str)
	local ids = {}
	local _, n = str:gsub('%i', '')
	for i = 1, n do
		ids[i] = math.random(2^32)
	end
	return string.format(str, table.unpack(ids))
end

local content = generateContent([[<> <<a:#%i>> <<#%i><<@%i>><@!%i><@&%i<#%i<:name:%i>>><a:name:%i>>><@&%i>]])
```

Mehgugs:
using content like that i get this:
```Overall there are 40815 mentions within the buffer
buffer_size: 1267635
Performing 500 tests of each method
Finished collecting data
Avg lpeg speed over 500 tests: 17.029600ms
Avg gmatch speed over 500 tests: 45.923826ms
Avg rex speed over 500 tests: 45.220727ms```

SinisterRectus:
Unsurprisingly, I decided to go with my own method for fixing this issue with ad69aed791a88ff5084a8bfbd165fa189cb10bf2. Although a single Lua pattern or fancy lpeg patterns are impressive, I think they are unnecessarily complicated and difficult to maintain. I went with something more modular instead. Performance does take a hit due to the multiple `gmatch` loops, but it's still efficient in that it will only look for mentions when the user accesses the individual properties instead of all at once. Sorry for rejecting all of your hard work.

PR #122, opened 2018-05-05
Lpeg parse mentions

Mehgugs:
Adds a new parseMention function which can correctly handle all types of mentions, it behaves the same as the previous parseMention; returning arrays of ids as output.

As discussed in  PR [#121](https://github.com/SinisterRectus/Discordia/pull/121) an lpeg based parser seems to be the fastest method to parse mentions accurately. The advantages to lpeg when compared to lrexlib and lua patterns are it's speed and accuracy. 

- The lpeg parser does not need to do any extra filtering and inserts mentions directly into the output arrays as they are found whilst checking for duplicate ids.
- The predicate patterns allow for a quick check before entering the larger pattern groups, this allows it quit parsing early.
- Easier to expand if discord adds new mentionables; the existing patterns are all independent of eachother.

```
The mention string inserted randomly into 100000 bytes of lorem ipsum was: "<@!word:1234><a@!word1234><><<a:#1234>><##1234><a::name:1234><<#1234><<@1234>><@!1234><@&1234<#1234<:name:1234>>><a:name:12344>>>"
This contains 5 correct mentions
Overall there are 34265 mentions within the buffer
buffer_size: 984363
Performing 1000 tests of each method
Finished collecting data
Avg lpeg speed over 1000 tests: 16.633165ms
Avg gmatch speed over 1000 tests: 36.292022ms
Avg rex speed over 1000 tests: 36.292862ms
```

### DIFF
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -10,7 +10,7 @@ local insert = table.insert
 local null = json.null
 
 local lpeg = require"lpeg"
-local P, V, C, Ct, S, Carg, Cc, l = lpeg.P, lpeg.V, lpeg.C, lpeg.Ct, lpeg.S, lpeg.Carg, lpeg.Cc, {} 
+local P, V, C, S, Carg, l = lpeg.P, lpeg.V, lpeg.C, lpeg.S, lpeg.Carg, {} 
 lpeg.locale(l)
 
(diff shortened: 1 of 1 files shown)


SinisterRectus:
See #121

PR #128, opened 2018-05-17
doc strings

truemedian:
descriptions soon™

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -258,7 +258,7 @@ end
 --[=[
 @m run
 @p token string
-@p presence table
+@op presence table
 @r void
 @d ...
 ]=]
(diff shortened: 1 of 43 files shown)


SinisterRectus:
Can you change the "void" returns to "nil"? I know there is a technical difference between returning nil and returning nothing in LuaJIT, but void isn't an actual type, and:

```lua
local function a() return end
local function b() return nil end
print(a() == b()) -- true

SinisterRectus:
... or just don't include a return type when there is no explicit return.

SinisterRectus:
Looks good from what I can see. I'll probably make some changes as I continue to go through it all. Thanks for all the help @truemedian @PurgePJ @Tigerism

PR #129, opened 2018-05-21
Changed emitted event(s) from MESSAGE_DELETE_BULK

PurgePJ:
Now emits 'messageDeleteBulk' with a table of message objects (or a basic {id = id, channel = channel, uncached = true} table for those uncached messages).

Simple example:
```lua
client:on("messageDeleteBulk", function(messages)
	print(#messages .. " messages deleted.")

	for _, message in pairs(messages) do
		p("Message author was: " .. message.author.username)
		p("Message ID: " .. message.id)
		p("Message content was : " .. message.content)
		p()
	end
end)
```

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -89,7 +89,7 @@ end
 local Client, get = require('class')('Client', Emitter)
 
 function Client:__init(options)
-	Emitter.__init(self)
+	self._emitter = Emitter.__init(self)
 	options = parseOptions(options)
 	self._options = options
 	self._shards = {}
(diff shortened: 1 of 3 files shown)


SinisterRectus:
This is a breaking change. While it's not an explicit change like removing a class method would be, this does change the expected behavior of the library. Users expect bulk-deleted messages to appear as regular `messageDelete` events. We should not suddenly change that.

PurgePJ:
We will move this for Discordia 3.x.

Please close this PR.

SinisterRectus:
I think authors can close PRs. If not, you can always just remove the branch.

PR #134, opened 2018-07-13
4003 after outage

truemedian:
return error for websocket messages that are sent in the time frame between invalid session and re-identify that trigger 4003 errors from the socket.


SinisterRectus:
Thank you.

SinisterRectus:
This prevents IDENTIFY from being sent because there is no session_id on start up. Please fix ASAP or we can just revert it.

PR #135, opened 2018-07-16
Fix of fix for 4003 issues

Mehgugs:
Identical behaviour to the previous fix except the bug which prevented `IDENTIFY` from sending is patched, and the `WebSocket` class is left untouched. This moves the fix for a discord issue into the discord specific `Shard` class instead.

It may be preferable to check the gateway OP code but I think the extra argument is a bit more flexible should more issues arrive in the future.

### DIFF
--- a/libs/client/Shard.lua
+++ b/libs/client/Shard.lua
@@ -50,6 +50,16 @@ function Shard:__init(id, client)
 	self._backoff = 1000
 end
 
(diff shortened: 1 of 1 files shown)


SinisterRectus:
ef439fc7949c9ee8b8bee406550ac3171cce5efc

PR #137, opened 2018-07-26
Fixed client

pmx625:
This fixes the following error:
discordia/libs/voice/VoiceConnection.lua:312: attempt to index field '_client' (a nil value)


pmx625:
Nevermind, it was something on my end that was causing the error. It is fixed now.
Thank you.

PR #138, opened 2018-08-10
Fixed predicated Emitter:waitFor

Mehgugs:
Currently Emitter:waitFor will only process one event. This does not make sense when you have a predicate which should filter events until one matches.

### DIFF
--- a/libs/utils/Emitter.lua
+++ b/libs/utils/Emitter.lua
@@ -195,11 +195,13 @@ provided, events that do not pass the predicate will be ignored.
 ]=]
 function Emitter:waitFor(name, timeout, predicate)
 	local thread = running()
-	local fn = self:onceSync(name, function(...)
+	local fn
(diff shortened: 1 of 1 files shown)


SinisterRectus:
I'm not convinced that there is an issue, or that this does not introduce one. Can you elaborate?

Mehgugs:
```lua
client:waitFor("messageCreate", 3e4, function(m) return m.author.id == ID end)
```
#### What the code looks like it does: 
Resumes with the first event to satisfy predicate within the timeout. 

#### What it actually does:
Waits for one event. If predicate succeeds then it resumes otherwise it just times out.
So the example above checks if the next messageCreate matches the predicate or times out. 

The PR changes it to work like the first case i described.

SinisterRectus:
I didn't realize that `onceSync` was the original code on my phone. It's obvious on desktop. Thanks for the fix.

PR #143, opened 2018-09-04
Add Guild:getBan(user)

PurgePJ:
This PR adds the missing function ``Guild:getBan(user)``.

It accepts both, a user ID or a user object.

### DIFF
--- a/libs/client/API.lua
+++ b/libs/client/API.lua
@@ -472,13 +472,13 @@ function API:removeGuildMember(guild_id, user_id, query) -- Guild:kickUser
 	return self:request("DELETE", endpoint, nil, query)
 end
 
-function API:getGuildBan(guild_id, user_id) -- Guild:getBan
-	local endpoint = f(endpoints.GUILD_BAN, guild_id, user_id)
+function API:getGuildBans(guild_id) -- Guild:getBans
+	local endpoint = f(endpoints.GUILD_BANS, guild_id)
 	return self:request("GET", endpoint)
 end
 
-function API:getGuildBans(guild_id) -- Guild:getBans
-	local endpoint = f(endpoints.GUILD_BANS, guild_id)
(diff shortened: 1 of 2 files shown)


libs/client/API.lua
```diff
@@ -472,6 +472,11 @@ function API:removeGuildMember(guild_id, user_id, query) -- Guild:kickUser
 	return self:request("DELETE", endpoint, nil, query)
 end
 
+function API:getGuildBan(guild_id, user_id) -- Guild:getBan
```
SinisterRectus: This should be below `getGuildBans`, as it is on the Discord docs, and as it is generated by my docsdump file.

libs/containers/Guild.lua
```diff
@@ -424,6 +424,25 @@ function Guild:pruneMembers(days)
 	end
 end
 
+--[=[
+@m getBan
+@p user User-ID-Resolvable | User
+@r Ban
+@d Returns a [ban object](https://discordapp.com/developers/docs/resources/guild#ban-object) for the given user.
```
SinisterRectus: Do not link to the Discord docs in our docstrings. A Discordia object is not the same as a Discord API JSON object. Also the link can break without warning.

SinisterRectus: Also say something about how the object may not exist.

libs/containers/Guild.lua
```diff
@@ -424,6 +424,25 @@ function Guild:pruneMembers(days)
...
+]=]
+function Guild:getBan(user)
+	if user.id then
```
SinisterRectus: This is not how we do input resolution. See `Guild:getMember` for inspiration.

PurgePJ:
👍
There was a doc typo with getRole being getMember in Guild. You might want to regenerate docs for it.

SinisterRectus:
Thanks.

PR #158, opened 2018-12-06
breaking prefix concatenation

truemedian:
Fix the redundant and breaking concatenation of prefix in helpCommandExample.lua:28, as args[1] already contains the prefix


SinisterRectus:
Thanks.

PR #165, opened 2019-01-08
fix docgen, fix typos in docstrings

truemedian:


truemedian:
it fixes how multiple returns are handled, so they can be rendered properly

libs/containers/AuditLogEntry.lua
```diff
@@ -181,7 +174,7 @@ end
 
 --[=[
 @m getMember
-@r Member
+@r Member/nil
```
Bilal2453, line 177: What is this for?   This is *the* only return in Discordia that is documented with an explicit `nil`.

PR #174, opened 2019-02-19
Fixed Login Issue

FHGDev:
Fixed issue where users have to enter "Bot " before their token to log in.

### DIFF
--- a/libs/client/API.lua
+++ b/libs/client/API.lua
@@ -118,7 +118,7 @@ end
 
 function API:authenticate(token)
 	self._headers = {
-		{'Authorization', token},
+		{'Authorization', 'Bot '..token},
 		{'User-Agent', USER_AGENT},
 	}
 	return self:getCurrentUser()


SinisterRectus:
This would fix things only for users who are not prefixing their token. It would break for everyone who does.

I have no interest in automatic token prefixing at this time. Here is a copy paste from our Discord channel:

> discordia 0.x was written when bot accounts did not exist, so the prefix is irrelevant here.
> 
> at some point in the discordia 1.x history, bot accounts were added, and bot tokens had to be prefixed with "Bot " for HTTP requests. it is ignored for gateway connections. discordia 1.x did automatically prefix tokens with "Bot ". if the token did not work on gateway connections, it would try it without the prefix. if neither worked, it would exit.
> 
> in discordia 2.x, i added more helpful logging, which is enabled by default, and i also added a step during authentication that checks the validity of a token by using an HTTP request before connecting to the gateway. if i kept the same behavior from 1.x, every attempt to authenticate with a user token would make a failed request before a successful one. i did not want this failed request to be logged, or be made, with each login, so i left it up to the user to add the "Bot " when necessary, and omit it when it was not necessary. using user tokens was not legally forbidden by any developer ToS at this point.
> 
> eventually the developer ToS was released, and the general opinion of people here moved towards "don't use user tokens", which is an opinion that i have adapted, but for the purposes of consistency, i have not removed the relevant code from discordia. in version 3.x, i will probably remove it, and i will be able to automatically prefix tokens again.
> 
> i could have allowed users to explicitly tell the client that a user token is being used instead of a bot token (d.py does this or did this). i am not a fan of this, as i wanted authentication to be easy for users, and i thought that adding "Bot " to the token string, which is documented, was easier for users than it is for users to set a client option somehow, which is not documented.

PR #175, opened 2019-02-24
Add Snowflake.date

Mehgugs:
Adds `Snowflake.date`. This makes computation with "created ats" much more terse if desired and doesnt break the current behaviour. 

#### Example Usage:
```lua
return (msg.date - another_date):toString()
```

### DIFF
--- a/libs/containers/abstract/Snowflake.lua
+++ b/libs/containers/abstract/Snowflake.lua
@@ -35,6 +35,14 @@ function get.createdAt(self)
 	return Date.parseSnowflake(self._id)
 end
 
+--[=[@p date Date A date object representing the time at which the object was created by Discord.
+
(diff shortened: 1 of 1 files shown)


SinisterRectus:
This is, in my opnion, too much bloat, but I can see how it is useful.

Maybe a bigger issue is that I don't like properties getters that create and return a new reference type with each visit. Maybe it should be memoized.

SinisterRectus:
I think `Snowflake:getTimestamp()` would be more appropriate here.

We could add `Member:getJoinedAt()` and `Guild:getJoinedAt()` too.

SinisterRectus:
9bb49470de3e98da942ba90e26ff50ed2d9c5ee4

PR #183, opened 2019-04-11
Fix Member:hasPermission returning false in some conditions for admins

DrLynix:
Like if channel disallow everyone to see it, it would return false for the read permission, even for admins.

To reproduce the bug:
- Have a channel disallowing everyone from reading/writing in it
- Test if an admin has the read/write permission on that channel with Member:hasPermission

because channel overwrites permissions are checked before admin rights, this will return false (and should return true because admins do have all rights, well almost all rights).

### DIFF
--- a/libs/containers/Member.lua
+++ b/libs/containers/Member.lua
@@ -108,15 +108,15 @@ function Member:hasPermission(channel, perm)
 		return true
 	end
 
-	local rolePermission = guild.defaultRole.permissions
+	local rolePermissions = guild.defaultRole.permissions
 
 	for role in self.roles:iter() do
 		if role.id ~= guild.id then -- just in case
-			rolePermission = bor(rolePermission, role.permissions)
+			rolePermissions = bor(rolePermissions, role.permissions)
 		end
 	end
 
-	if band(rolePermission, permission.administrator) > 0 then
+	if has(rolePermissions, permission.administrator) then
 		return true
 	end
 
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Before I review this myself, can you just make sure that its behavior is consistent with https://discordapp.com/developers/docs/topics/permissions#permission-overwrites

DrLynix:
I think it needs more work to be fully consistent with what Discord does (like handling implicit permissions, `Member:hasPermission(channel, permission.attachFiles)` should return false if viewing the channel or writing to it is denied), however this fix is consistent toward Discord behavior and the  `Member:getPermissions` function.

All I did was taking the role permissions computation from `Member:getPermissions`, doing it before permissions overwrite, checking the admin flag (early returning true if set) then doing the permissions overwrite check as before if a channel if set and then only checking the role permissions.

It shouldn't change the function behavior in any way except for the bug described in the first message.

libs/containers/Member.lua
```diff
@@ -108,6 +108,18 @@ function Member:hasPermission(channel, perm)
...
 
+	local rolePermission = guild.defaultRole.permissions
+
+	for role in self.roles:iter() do
+		if role.id ~= guild.id then -- just in case
+			rolePermission = bor(rolePermission, role.permissions)
+		end
+	end
+
+	if band(rolePermission, permission.administrator) > 0 then
```
SinisterRectus: Change this to use the `has` helper.

DrLynix:
Here you are

PR #185, opened 2019-05-11
Made emoji capture less greedy

leonardus:
Discord emojis are 2+ characters long with alphanumeric characters and underscores. The `.+` capture resulted in unintended consequences when there was more than one emoji in the message.

### DIFF
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -400,7 +400,7 @@ function get.cleanContent(self)
 			:gsub('<@!?(%d+)>', users)
 			:gsub('<@&(%d+)>', roles)
 			:gsub('<#(%d+)>', channels)
-			:gsub('<a?:[%w_]+:(%d+)>', '%1')
+			:gsub('<a?(:[%w_]+:)%d+>', '%1')
 			:gsub('@everyone', everyone)
 			:gsub('@here', here)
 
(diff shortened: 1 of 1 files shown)


leonardus:
We could use the `<a?(:[%w_]+:)(%d+)>` pattern to be consistent with the rest of the file, or use `<a?(:[%w_][%w_]+:)(%d+)>` to be consistent with Discord's specs for names of server emojis, which requires a minimum of two characters.

SinisterRectus:
Can you give an example string that causes the original issue and show how this change fixes that issue? Thank you.

leonardus:
`("<:emoji:12345><:emoji:12345><:emoji:12345><:emoji:12345>"):gsub('<a?(:.+:)%d+>', '%1')`
returns:
`:emoji:12345><:emoji:12345><:emoji:12345><:emoji:	1`

leonardus:
The new pattern would return the expected output of `:emoji::emoji::emoji::emoji:	4`.

libs/containers/Message.lua
```diff
@@ -400,7 +400,7 @@ function get.cleanContent(self)
 			:gsub('<@!?(%d+)>', users)
 			:gsub('<@&(%d+)>', roles)
 			:gsub('<#(%d+)>', channels)
-			:gsub('<a?(:.+:)%d+>', '%1')
+			:gsub('<a?:[%w_]+:(%d+)>', '%1')
```
truemedian:
```suggestion
			:gsub('<a?(:[%w_]+:)%d+>', '%1')
```
This needs to be changed to capture the actual name of the emoji instead of the id, this is correct behavior for cleanContent

PR #209, opened 2019-09-14
Add ability to customize gateway file location;

Playermet:


SinisterRectus:
Is there a reason you want to control the location? This file is intended to be used only by the library, not by users.

Playermet:
I just want to keep my application root directory clean, so I trying move all temp/cache/etc. files to other directory.

Issue #142, opened 2018-09-02
Discordia's caching model 


Mehgugs:
Currently discordia objects store there children in an instance of the `Cache` class (or it's derivatives) however the `Cache` class is used as a container for an object's children **and** a object cache simultaneously.

The issue with a hybrid model like this is that there has to be a trade off between retaining state so that events can have good context (`messageDelete` being aware of the message's properties for example.) and also reflecting changes from HTTP requests et al. instantly. (The count of `channel.messages` should always decrement when you delete one of it's messages.)

My suggestion is that discordia separates these two concerns into different objects; so that the logic for both remains simple. (Trying to handle the edge cases in one object is a problem.)

Add a new `Collection` class which discordia objects use to store their children, calling methods on this class to change it's contents happen synchronously so it doesn't need to try and keep hold of objects.

Change the `Cache` class to be a cache of full objects which should keep objects until they are known to be removed. This class would only be focused on storing objects not on reflecting events immediately.

SinisterRectus:
This will almost certainly not change until the next major version of Discordia.

SinisterRectus:
This has been addressed in commit 4eefacad005903f6768654f0fbd53da9a7535e0f. I'll keep this idea in mind, but I'd like to stick with a simpler solution moving forward rather than introduce more classes. I think that the next step for improving message caching would be that which is outlined in #214.

SinisterRectus:
Decided to re-open this because the idea here is good. Not yet sure if it will be implemented though.

SinisterRectus:
Caching has been overhauled for 3.0. Caches and cached objects are now hidden internally in a central client state manager. If the client requests permanently cached objects, for example, a guild's roles, the client collects the objects from the internal cache and returns references to them in a newly constructed static iterable. The occurrence of objects in that iterable does not change (unless the user manipulates the iterable). For example, if a role is deleted on Discord's server, it will be removed from the client's internal cache, but the role will still exist in the static iterable as long as the user holds onto it. The objects themselves may change as updates are received for them, since they are references to the same objects stored and manipulated in cache.

PR #212, opened 2019-09-30
Implement guildSubscriptions for bots.

NCPlayz:
This allows you to opt out of guild update events, such as `presenceUpdate`. More information is available at discordapp/discord-api-docs#1016.

```lua
local client = discordia.Client {guildSubscriptions = true}

client:on("presenceUpdate", function(member)
	print(member) -- won't print anything because it will never receive a presence update
end)
```

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -67,6 +67,7 @@ local defaultOptions = {
 	gatewayFile = 'gateway.json',
 	dateTime = '%F %T',
 	syncGuilds = false,
+	guildSubscriptions = true,
 }
 
 local function parseOptions(customOptions)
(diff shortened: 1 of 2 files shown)


SinisterRectus:
Superficially adding an option for guild subscriptions is okay because it reduces network throughput, but this misses a major point of why the feature exists. We need to adjust how Discordia caches members and presences while guild subscriptions are disabled so that the objects don't become stale.

NCPlayz:
Okay, I will definitely rework this PR since I know I didn't put any effort into it. Hopefully I rework the caching behaviour by the end of this week.

SinisterRectus:
Yeah I noticed.

SinisterRectus:
Implementing this relies on changes to Discordia's caching, which won't occur until I finish a rewrite. See #214 and #142

PR #261, opened 2020-09-24
Correct mistake in documentation for `Permissions`

dekkonot:


SinisterRectus:
Thanks for the fix. I will merge it, but since this is a minor concern, I probably won't rebuild the docs until later.

Issue #140, opened 2018-08-21
Activity has no hash


SinisterRectus:
Vaguely related to #139, `Activity` has no unique `__hash` method; it uses the default, which just returns `self`. There is no definitive way to hash activities, though.

SinisterRectus:
This will almost certainly not change until the next major version of Discordia.

PR #265, opened 2020-10-05
Add an implementation for Activity.__hash

Dekkonot:
As a consequence of issue #140, `tostring(Activity)` (and `print(Activity)` as a result) causes Luvit to exit without an error code or stacktrace. Your guess as to why it crashes is as good as mine, but given the nature of the crash, it seems like it would be a good idea to add a `__hash` implementation now rather than waiting for 3.0.

This implementation optimistically assumes there will only be one activity type per member at a time, which as far as I know should be the case. If that's not the case, let me know and I'll close this PR.

### DIFF
--- a/libs/containers/Activity.lua
+++ b/libs/containers/Activity.lua
@@ -39,6 +39,11 @@ function Activity:_loadMore(data)
 	self._emoji_animated = emoji and emoji.animated
 end
 
+--[=[
+@m __hash
+@r string
+@d Returns `Activity.type .. Activity.parent:__hash()`
+]=]
 function Activity:__hash()
 	return format("%i%s", self._type, self._parent:__hash())
 end
--- a/libs/containers/Activity.lua
+++ b/libs/containers/Activity.lua
@@ -42,10 +42,10 @@ end
 --[=[
 @m __hash
 @r string
-@d Returns `Activity.type .. Activity.parent:__hash()`
+@d Returns `Activity.parent:__hash()`
 ]=]
 function Activity:__hash()
-	return format("%i%s", self._type, self._parent:__hash())
+	return self._parent:__hash()
 end
 
 --[=[@p start number/nil The Unix timestamp for when this Rich Presence activity was started.]=]
--- a/libs/containers/Activity.lua
+++ b/libs/containers/Activity.lua
@@ -39,6 +39,10 @@ function Activity:_loadMore(data)
 	self._emoji_animated = emoji and emoji.animated
 end
 
+function Activity:__hash()
+	return format("%i%s", self._type, self._parent:__hash())
+end
+
 --[=[@p start number/nil The Unix timestamp for when this Rich Presence activity was started.]=]
 function get.start(self)
 	return self._start


SinisterRectus:
The `__hash` methods in Discordia currently have three uses:

- So that you can have `cache[obj:__hash()] = obj` as a general form for caching different containers.
- The  `__eq` method compares the hash of each container.
- The `__string` method uses the hash to build the container's string.

I never added the method for activities because they are not stored in actual caches. That this would affect the `__eq` and `__tostring` metamethods was an oversight on my part.

I honestly didn't know about the string bug. You are indeed seeing a stack overflow due to the missing method. This is a bug for sure (hence issue #140), but I wouldn't consider it critical only because it's existed for so long and no one noticed until now.

----

Regarding your fix, can you explain why you're using the `type` property. As far as I can tell, the type does not discriminate the activity. Maybe I'm missing something?

I want to say it's sufficient to just use the activity's parent's hash, but it may not be and that's because the parent's hash may also not be sufficient as far as hashes go. The parent is the member object, and a member's hash is only `member.user.id`. In Discordia's current cache setup, that is sufficient enough to distinguish members when they are deliberately cached per guild., but if you ever wanted to cache members in one central location, not per guild, you will have a hash collision for member objects that belong to the same user on different guilds. A hash should be more unique in that case.

When considering the equality of two activities, should every field be compared? If all fields are equal, are those two activities equal, or must the members also be equal, too? And by what definition should the members be equal? Should the member's guild be considered?

Additionally, how useful do you want the activity's string to be? Should it dump all the information about the activity?

Overall, it's an interesting issue, but I haven't bothered to fix it because it is complex. In 3.0, I have not defined any `__hash` methods. Instead, I am uniquely defining each `__eq` and `__tostring` method, and I am explicitly indexing the necessary properties during caching. For example `cache[member.guildId][member.user.id] = member`. I don't yet have an activity object. I do have a presence object, but I have not defined any `__eq` or `__tostring` methods for these. Still plenty of work to get done.

----

tl;dr: This is a complicated issue and I'm not sure whether I want to even touch it. However, why are you using the activity type?

Dekkonot:
I was using the activity type based on the assumption that there was only ever going to be one activity per type per member at any given time. The logic being that there's really only 3-ish activity types (statuses, game, and streaming) so it would be reasonable to assume they couldn't all happen at once and that the activity type would be enough to differentiate between them. After some more thorough testing though, that's wrong, so that's my bad. Apologies.

With regards to actually solving the issue, it might actually be enough to have the activities share their member's ID; seems like a similiar scenario to bans (where the hash points to the user's ID) or even members for the moment. Not very useful for caching globally but for the time being it would work. The goal shouldn't be to differentiate between members and activities but to differentiate between activities. Which, at the moment, my PR fails to do because apparently `type` is `nil` for custom statuses in the same way it is for no status. Thanks Discord.

Comparing equality between two activities does present an interesting problem but I'm not sure if it's that complicated, just tedious -- you would ideally compare different fields based on what the activity was to determine if they were equal. Implementation details aside, that seems the most useful to me as a user since I probably don't care if someone's gamemode or party has changed so long as they're still playing the same game.

I suppose if you're doing away with Hashes entirely in 3.0, it might not be worth the effort to come up with a good solution in 2.X. If that's the case, would you be content with simply making the hash the member's ID? It would solve the crashing issue and give an identifiable hash to an Activity in the same way things like Bans and Relationships do.

SinisterRectus:
The parent ID should be good enough for now.

Dekkonot:
Done.

PR #266, opened 2020-10-06
fix: Invite.createdAt and Webhook:getDefaultAvatarURL

truemedian:
* Invite.createdAt used `self._createdAt` instead of `self._created_at`

* Webhook:getDefaultAvatarURL was missing a parameter


SinisterRectus:
You already fixed these in #263.

truemedian:
I'm really blind then sorry, I must have forgotten to pull the fork to master after that merge.

PR #274, opened 2020-12-23
allow sorter function to be passed to toArray

GitSparTV:
Completes https://github.com/SinisterRectus/Discordia/projects/2#card-34095190

### DIFF
--- a/libs/utils/Bitfield.lua
+++ b/libs/utils/Bitfield.lua
@@ -127,7 +127,7 @@ function Bitfield:toArray(filter, sort)
 			insert(arr, k)
 		end
 	end
-	if sort and typing.checkType('table', sort) then
+	if sort and typing.checkType('function', sort) then
 		table.sort(arr, sort)
 	end
 	return arr
--- a/libs/utils/Bitfield.lua
+++ b/libs/utils/Bitfield.lua
@@ -120,12 +120,15 @@ function Bitfield:__div(other)
 	end
 end
 
-function Bitfield:toArray(filter)
+function Bitfield:toArray(filter, sort)
 	local arr = {}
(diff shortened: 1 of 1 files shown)


libs/utils/Bitfield.lua
```diff
@@ -120,13 +120,16 @@ function Bitfield:__div(other)
...
 
-function Bitfield:toArray(filter)
+function Bitfield:toArray(filter, sort)
 	local arr = {}
 	for k, v in pairs(checkType('table', filter)) do
 		if self:hasValue(v) then
 			insert(arr, k)
 		end
 	end
+	if sort and typing.checkType('table', sort) then
```
truemedian, line 130: `table` is the wrong type here. It must be a `function`.

GitSparTV: copy-paste mistake, thanks

Bilal2453: `sort` cannot be false, should be `if sort ~= nil` instead if I'm not mistaken

truemedian: type checking covers any other case

GitSparTV: If this is required, alright.

truemedian:
Also note, this is only applied to Bitfield at the moment, eventually it will need to be added to any iterable object.

GitSparTV:
Sinister, should `false` error or not?

SinisterRectus:
The project card is actually for the `Iterable:toArray` method that I have not yet implemented. I'm not sure that I want to include sorters though. I think it's fairly trivial for a user to use `table.sort` on the result if necessary, avoiding the complexity of including it in the method.

PR #275, opened 2020-12-24
Fix Date.fromISO not reading timezone

GitSparTV:
Doesn't support minutes offset, but at least it reads hours and doesn't error on Discord timestamps like `guild.joined_at`.

### DIFF
--- a/libs/utils/Date.lua
+++ b/libs/utils/Date.lua
@@ -44,6 +44,7 @@ local properties = { -- name, pattern, default
 	{'min', '%d%d%d%d%-%d%d%-%d%d.%d%d:(%d%d)', 0},
 	{'sec', '%d%d%d%d%-%d%d%-%d%d.%d%d:%d%d:(%d%d)', 0},
 	{'usec', '%d%d%d%d%-%d%d%-%d%d.%d%d:%d%d:%d%d.(%d%d%d%d%d%d)', 0},
+	{'zone', '%d%d%d%d%-%d%d%-%d%d.%d%d:%d%d:%d%d.%d%d%d%d%d%d([+-]%d%d):%d%d', 0}
 }
 
 local function toTime(tbl, utc)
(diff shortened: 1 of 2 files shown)


SinisterRectus:
As far as I know, an ISO string is always +00:00; at least it is in Discord, which is the most important context here. Is there need or precedent for including different time zones in the string?

GitSparTV:
Yes, try to read guild.created_at using Date.fromISO() without my fix

GitSparTV:
You don't need to use it but you have to read it

SinisterRectus:
Why?

GitSparTV:
I didn't find the way to ignore zone and make Discord timestamps work

SinisterRectus:
I think that if we want to support time zone offsets, we need to handle an arbitrary suffix. For every accepted pattern, there can be a time zone version.

For example, if these are accepted:
```
YYYY
YYYY-MM
YYYY-MM-DD
```
then these should be too:
```
YYYY+/-nn:nn
YYYY-MM+/-nn:nn
YYYY-MM-DD+/-nn:nn
```

Following that, if we read time zones, then I think that the Date class should use them consistently, but that adds complexity. For example, what would `Date.fromISO(str):toISO()` return if `str` has a time zone offset? Do you have to keep a time zone state on the Date object? Do you just apply the offset to the time and force the string to 00:00?

GitSparTV:
This PR is just for fixing discord timestamps. Only thing I can suggest is to match but not read last element of full iso timestamp

SinisterRectus:
In that case, this is not the correct fix. The problem is in this line `if not valid and i == 1 and j == #str then`. It's assuming that the full string is being matched with `j == #str`.

GitSparTV:
Yes

SinisterRectus:
I think we can either ignore everything after the final capture, or adjust the capture to ignore the time zones.

SinisterRectus:
I'm trying to fix this, but I'm finding that it's not as easy as I thought it would be. It might be worth to time to add some proper timezone offset handling.

SinisterRectus:
See c164fb72772b97dd7129bf9bc5c2726c3ad310d8.

PR #276, opened 2020-12-24
Optimizations

GitSparTV:
Draft PR.

- Optimized __index and __pairs in enums


SinisterRectus:
I purposefully use closures for these to prevent the original table from being exposed.

PR #278, opened 2021-01-06
Relay allowed_mentions table

Python1320:


Python1320:
It is useful to prevent mentions or am I wrong? It seems to work. 
Test with:  https://github.com/Metastruct/metaluvit/blob/a531675f9ca85e6900e65418a0c8ef687381b89b/autorun/command_lua.lua#L42

SinisterRectus:
If you'd like to maintain your own version of Discordia that has this addition, that's fine, but I cannot accept it here because there is no attempt made to integrate the feature with Discordia's API.

I will add support for this in 3.0.

PR #287, opened 2021-04-12
Specifies the use of a bot token.

OneFoggyScreen:
Discord has changed it to where in authorization needs a differential between bot and bearer tokens. This fix allows for the bot to again access the API without errors.

### DIFF
--- a/libs/client/API.lua
+++ b/libs/client/API.lua
@@ -118,7 +118,7 @@ function API:__init(client)
 end
 
 function API:authenticate(token)
-	self._token = token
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Discord hasn't changed anything here. Discordia just doesn't automatically prefix the token. You have to do it manually when you input your token in the `client:run` method. 

In the future, in Discordia 3.0, there will be a `tokenPrefix` client option that will default to `Bot ` so that users do not need to manually prefix their tokens.

PR #293, opened 2021-05-19
support replies

IsLilyYaGirl:
support message replies

### DIFF
--- a/libs/containers/Message.lua
+++ b/libs/containers/Message.lua
@@ -357,10 +357,10 @@ end
 @t http
 @p content string/table
 @r Message
-@d Equivalent to `Message.channel:send(content)`.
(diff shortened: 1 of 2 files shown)


IsLilyYaGirl:
had this in my personal fork and thought of making a pull request for it

IsLilyYaGirl:
yes i did test it

SinisterRectus:
Hi, thank you for this. The problem with your implementation is that all calls to `message:reply` will include an actual reply with a mention by default. This isn't exactly a breaking change, but it does change the expected behavior of the library, so we should not do this.

I've gone and written my own implementation and opened #294 for people to comment.

IsLilyYaGirl:
i see, thanks

PR #296, opened 2021-05-31
fix mentionedUsers error on reply with no mentions

truemedian:
mentionedUsers uses `parseMentions` to obtain a table of mentioned users in the message and then adds the reply target to that table if the message is a reply. This fails when the message contains no mentions (fails `%b<>`) and the message is a reply. This fixes the issue by having `parseMentions` return an empty table in such the case.


SinisterRectus:
Not the fix I would use, but it's the easiest for now. Thanks.

PR #298, opened 2021-06-27
Update components

borsuczyna:
Hey, I simply updated Message:update code to update components, I think you should add this: 
```lua
function Message:update(data)
	return self:_modify({
		content = data.content or null,
		embed = data.embed or null,
		components = data.components or null,
	})
end

greetings!

### DIFF
--- a/libs/client/EventHandler.lua
+++ b/libs/client/EventHandler.lua
@@ -545,9 +545,10 @@ function EventHandler.INTERACTION_CREATE(d, client)
 	http.request("POST", url, headers, payload)
     	local guild = client._guilds:get(d.guild_id)
 	local channel = getChannel(client, d.channel_id)
+	local buttonid = d.data.custom_id
     	local member = d.member and guild._members:_insert(d.member) or guild._members:get(d.user_id)
 	local message = channel._messages:_insert(d.message)
-    	return client:emit('buttonPressed', member, message)
+    	return client:emit('buttonPressed', buttonid, member, message)
 end
 
 return EventHandler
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: libs/client/EventHandler.lua [hunk 1]
```diff
@@ -537,4 +537,18 @@ function EventHandler.WEBHOOKS_UPDATE(d, client) -- webhook object is not provid
...
+	local payload = "{\"type\": 6}"
+	local http = require("coro-http")
+	http.request("POST", url, headers, payload)
+    	local guild = client._guilds:get(d.guild_id)
+	local channel = getChannel(client, d.channel_id)
+	local buttonid = d.data.custom_id
+    	local member = d.member and guild._members:_insert(d.member) or guild._members:get(d.user_id)
+	local message = channel._messages:_insert(d.message)
+    	return client:emit('buttonPressed', buttonid, member, message)
+end
```
truemedian, line 552: The indentation here is inconsistent with the rest of discordia, it should be indented with a single tab.

libs/client/EventHandler.lua
```diff
@@ -537,4 +537,18 @@ function EventHandler.WEBHOOKS_UPDATE(d, client) -- webhook object is not provid
 	return client:emit('webhooksUpdate', channel)
 end
 
+function EventHandler.INTERACTION_CREATE(d, client)
+	local url = "https://discord.com/api/v8/interactions/"..d.id.."/"..d.token.."/callback"
+	local headers = {{"Content-Type", "application/json"}}
+	local payload = "{\"type\": 6}"
+	local http = require("coro-http")
+	http.request("POST", url, headers, payload)
```
truemedian, line 545: Making a HTTP request behind the scenes the exact kind of scenario that discordia goes out of it's way to avoid. This shouldn't be here.

truemedian (changes requested):
This PR entirely clobbers the README, it shouldn't change it at all. The commit history also a mess that really needs to be rebased into a single commit.

callmehSpear:
Man literally changed the readme file-

PR #305, opened 2021-08-07
class.mixin returns the target table

RiskoZoSlovenska:
Currently, `class.mixin()` returns no values. By having it return the `target` argument, the code
```lua
local myMixin = {}
class.mixin(myMixin, someOtherMixin)
```
can be rewritten much more neatly as
```lua
local myMixin = class.mixin({}, someOtherMixin)
```


SinisterRectus:
Function should probably do some type checking, but this is fine.

PR #311, opened 2021-10-01
Update package.lua

Perthys:


SinisterRectus:
Shame on me for never updating this. I guess it's okay to update now, but it will need it again in just 3 months.

Note that you have not updated the date on the LICENSE file.

Perthys:
updated the license too

PR #318, opened 2021-10-13
Update message.lua (Prevent bot from crashing when a sticker or wave is sent)

Richy-Z:
Discord added a new feature called "wave", where a user gets to send a waving sticker when someone else joins a guild. When that wave is sent, it causes the bot to spontaneously crash because Discordia didn't know who the wave mentioned. Pretty sure that this is the solution because I tried it and it stopped my bot from crashing when the wave is sent.


Richy-Z:
Thanks, didn't realise that. I'll close the pull request.

PR #332, opened 2022-01-15
Member: implement timeouts

Bilal2453:
This will add 3 methods into Member class, `:timeoutFor(duration)`/`:timeoutUntil(Date)`/`:removeTimeout()`. And a single boolean getter `.timedout` which uses the `communication_disabled_until` property as its state; this might make the getter a bit expensive to call since it has to create two Date objects and compare them, but it seems like this is the way Discord intend you to do it:

> when the user's timeout  will expire and the user will be able to communicate in the guild  again, null or a time in the past if the user is not timed out

I've tested the changes a bit, everything seem to work as expected, the cache properly gets updated for the state tracking. The doc-comments might need a bit of language review though to make sure it is clear enough.

### DIFF
--- a/libs/containers/Member.lua
+++ b/libs/containers/Member.lua
@@ -15,6 +15,8 @@ local Color = require('utils/Color')
 local Resolver = require('client/Resolver')
 local GuildChannel = require('containers/abstract/GuildChannel')
 local Permissions = require('utils/Permissions')
+local Date = require('utils/Date')
+local Time = require('utils/Time')
 
 local insert, remove, sort = table.insert, table.remove, table.sort
 local band, bor, bnot = bit.band, bit.bor, bit.bnot
@@ -464,6 +466,60 @@ function Member:unban(reason)
 	return self._parent:unbanUser(self._user, reason)
 end
 
+function Member:_timeout(val)
+	local data, err = self.client._api:modifyGuildMember(self._parent._id, self.id, {communication_disabled_until = val or json.null})
+	if data then
+		self._communication_disabled_until = val ~= json.null and val or nil
+		return true
+	else
+		return false, err
+	end
+end
+
+--[=[
+@m timeoutFor
+@t http
+@p duration Time/number
+@r boolean
+@d Timeouts the member in its Guild.
+`duration` is either a Time object or a number of seconds representing the duration the timeout stays for.
+]=]
+function Member:timeoutFor(duration)
+	if type(duration) == 'number' then
+		duration = (Date() + Time.fromSeconds(duration)):toISO()
+	elseif isInstance(duration, Time) then
+		duration = (Date() + duration):toISO()
+	end
+	return self:_timeout(duration)
+end
+
+--[=[
+@m timeoutUntil
+@t http
+@p date Date/number
+@r boolean
+@d Timeouts the member in its Guild.
+`date` is either a Date object or a UNIX epoch in seconds at which the member's timeout ends.
+]=]
+function Member:timeoutUntil(date)
+	if type(date) == 'number' then
+		date = Date(date):toISO()
+	elseif isInstance(date, Date) then
+		date = date:toISO()
+	end
+	return self:_timeout(date)
+end
+
+--[=[
+@m removeTimeout
+@t http
+@r boolean
+@d Removes the timeout of the member.
+]=]
+function Member:removeTimeout()
+	return self:_timeout()
+end
+
 --[=[@p roles ArrayIterable An iterable array of guild roles that the member has. This does not explicitly
 include the default everyone role. Object order is not guaranteed.]=]
 function get.roles(self)
@@ -518,6 +574,16 @@ end
 function get.deafened(self)
 	local state = self._parent._voice_states[self:__hash()]
 	return state and (state.deaf or state.self_deaf) or self._deaf
+end
+
+--[=[@p timedout boolean Whether the member is timed out in its guild.]=]
+function get.timedout(self)
+	local state = self._communication_disabled_until
+	if not state then
+		return false
+	else
+		return Date.fromISO(state) > Date()
+	end
 end
 
 --[=[@p guild Guild The guild in which this member exists.]=]
--- a/libs/containers/Member.lua
+++ b/libs/containers/Member.lua
@@ -576,14 +576,20 @@ function get.deafened(self)
 	return state and (state.deaf or state.self_deaf) or self._deaf
 end
 
---[=[@p timedout boolean Whether the member is timed out in its guild.]=]
-function get.timedout(self)
+--[=[@p timedOut boolean Whether the member is timed out in its guild.]=]
+function get.timedOut(self)
 	local state = self._communication_disabled_until
 	if not state then
 		return false
 	else
 		return Date.fromISO(state) > Date()
 	end
+end
+
+--[=[@p timedOutUntil string/nil The raw communication_disabled_until member property.
+Note this may be provided even when the member's time out have expired.]=]
+function get.timedOutUntil(self)
+	return self._communication_disabled_until
 end
 
 --[=[@p guild Guild The guild in which this member exists.]=]
--- a/libs/containers/Member.lua
+++ b/libs/containers/Member.lua
@@ -481,8 +481,9 @@ end
 @t http
 @p duration Time/number
 @r boolean
-@d Timeouts the member in its Guild.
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Looks great. The Date and Time classes really do most of the heavy lifting here.

Is it worth changing `Member:_timeout` to a public `Member:setCommunicationDisabledUntil`?

Bilal2453:
> The Date and Time classes really do most of the heavy lifting here.

Yep, they came in handy. Probably a good reason to keep/implement them in Discordia 3 as well.

> Is it worth changing Member:_timeout to a public Member:setCommunicationDisabledUntil?

I cannot think of any scenario in which this can be useful for the end user. If they got an ISO string they can easily parse that into a Date object (does mean the performance is a bit affected since it will be converted into an ISO string again but that should be fine), without mentioning there is no good scenario in which the user input has to be ISO string.

Also because of that, I intentionally let any value acceptable, so if they want to do `:timeoutUntil("2022-01-15T16:48:51+00:00")` or `:timeoutFor(false)` (to remove the timeout) all be possible. So I don't really see that being much useful, therefore probably not worth it.

SinisterRectus:
What happens if you try to set a timeout for a member that is already times out?

Bilal2453:
> What happens if you try to set a timeout for a member that is already times out?

The timeout gets updated to the new one you set, for example, if a member have a 60 second timeout, and you call this again with 120 second, Discord will set a new timeout that's 120 second exactly.

If the user does not have a timeout, and you try to unset any timeout (that is, passing null) the request success without Discord doing anything.

RiskoZoSlovenska:
I can see someone wanting to get the date someone is timed-out until; perhaps it would be better to return a `Date` if the member is timed out, and `nil` if they aren't.

Bilal2453:
That's a good point, I haven't thought about that.  I don't know if it is a good idea for `timedout` to return anything other than a boolean, it kind of does not make sense. Perhaps we should implement a second getter? ~~.communicationDisabledUntil~~

RiskoZoSlovenska:
Why not just rename it to `timedoutUntil`? Users can still use the return value in an if statement, since a `Date` is truthy.

Bilal2453:
We are storing the state as an ISO string, we cannot store it as a Date object, because if for example the member is not cached, and a timeout was added (either by another bot or using the Discord client) the member cache will be updated with `communication_disabled_until` set to an ISO timestamp.  Therefor each call to `timedoutUntil` will basically be doing the following:
1. Take the input and convert it into a Date object.
2. Convert the date object into an ISO timestamp.
3. Convert the timestamp back into a Date object.
That sounds.. rather ridiculous to me; So another property may indeed make more sense.

RiskoZoSlovenska:
I don't really understand what you mean. Why not just do something like
```lua
function get.timedoutUntil(self)
	local state = self._communication_disabled_until
	if state then
		local untilDate = Date.fromISO(state)

		if untilDate > Date() then
			return untilDate
		end
	end
	return nil
end
```
which is not far at all from the current implementation.

SinisterRectus:
Just keep it as a string, like the joinedAt property.

Bilal2453:
> I don't really understand what you mean. Why not just do something like
> 
> ```lua
> function get.timedoutUntil(self)
> 	local state = self._communication_disabled_until
> 	if state then
> 		local untilDate = Date.fromISO(state)
> 
> 		if untilDate > Date() then
> 			return untilDate
> 		end
> 	end
> 	return nil
> end
> ```
> 
> which is not far at all from the current implementation.

This is exactly what I meant. This is the third conversion you have to be making here.
The first one is when a user uses `timeoutFor`, the input is converted into a Date object, parsed into ISO and sent. This takes the timestamp, parse it, and make a Date object for it.

Here:
```lua
function Member:timeoutFor(duration)
	if type(duration) == 'number' then
		duration = (Date() + Time.fromSeconds(duration)):toISO()
	elseif isInstance(duration, Time) then
		duration = (Date() + duration):toISO()
	end
	return self:_timeout(duration)
end
```
This creates the Date object and then parses it into iso timestamp to be sent (Discord only accept that). Then you use `timedoutUntil`, which reads the `_communication_disabled_until` state (that was already [set before](https://github.com/SinisterRectus/Discordia/pull/332/commits/af2470ae5c3267ed8e46a357e8089a1a466b8074#diff-8f19a4ca619de7034d374734c08df83583bb93627d88c3c1351e5d90a773919dR472) if the user had used `timeoutFor`) and then -again- converts it back into a Date object.

Basically what I am saying is that you are duplicating the steps.

> Just keep it as a string, like the joinedAt property.

Sounds like a good call, although should this be its own property is the question

SinisterRectus:
Oh, I didn't realize it wasn't... 

Does the API provide the string?

Bilal2453:
Oh no, Discord itself do provide the timestamp string of course. It is just that *if* the user had already used the method to set the timeout then tries to read `timedoutUntil` (which would return a Date object) we would be repeating ourselves.
image

Bilal2453:
So I guess we shall make `timedout` return the iso timestamp if the member was timedout (the timedout __can__ be an iso timestamp in the past if a timeout was set and then ended), if they were not we return a nil?

RiskoZoSlovenska:
Oh, you're assuming the user will be calling `timedoutUntil` right after timing a member out. In that case, why not make `:_timeout()` take a Date object, convert it to ISO and then return the Date object back?

RiskoZoSlovenska:
> So I guess we shall make timedout return the iso timestamp if the member was timedout (the timedout can be an iso timestamp in the past if a timeout was set and then ended), if they were not we return a nil?

For consistency's sake (with `joinedAt`, as Sinister mentioned), yeah, I think returning just the ISO string would be reasonable. In 3.0 though, it would be nice if both properties returned a `Date`.

Bilal2453:
I actually just realized we are exactly doing that right now huh, since Discord is sending a string timestamp, and the timestamp may be in the past if it was set then expired, I was already creating the Date object in `timedout` either ways. Right Risko, kind of forgot about that sorry.

```lua
function get.timedout(self)
	local state = self._communication_disabled_until
	if not state then
		return false
	else
		return Date.fromISO(state) > Date()
	end
end
```
The real issue would be just consistency

SinisterRectus:
Just return _communication_disabled_until as it is. Call it `timedOutUntil`. Uppercase O because timed out is two words.

Bilal2453:
Do we remove the current `timedout`? Just always returning the raw _communication_disabled_until is a bit of trouble for the user since they would have to compare it against the current date to know if the member was currently timed out or not

SinisterRectus:
It's convenient, but adds complexity. I can go either way. If you keep it, change it to `timedOut`.

Bilal2453:
This should be sufficient now. Do review the doc comment for timedOutUntil though

libs/containers/Member.lua
```diff
@@ -464,6 +466,60 @@ function Member:unban(reason)
...
+	end
+end
+
+--[=[
+@m timeoutFor
+@t http
+@p duration Time/number
+@r boolean
+@d Timeouts the member in its Guild.
+`duration` is either a Time object or a number of seconds representing the duration the timeout stays for.
```
SinisterRectus, line 485: Sets a timeout for a guild member. `duration` is either `Time` object or a `number` of seconds representing how long the timeout lasts. To set an expiration date, use `timeoutUntil` instead.

libs/containers/Member.lua
```diff
@@ -464,6 +466,60 @@ function Member:unban(reason)
...
+	return self:_timeout(duration)
+end
+
+--[=[
+@m timeoutUntil
+@t http
+@p date Date/number
+@r boolean
+@d Timeouts the member in its Guild.
+`date` is either a Date object or a UNIX epoch in seconds at which the member's timeout ends.
```
SinisterRectus, line 502: > Sets a timeout for a guild member. `date` is either `Date` object or a UNIX epoch in seconds at which the member's timeout ends. To set a duration, use `timeoutFor` instead.

SinisterRectus (changes requested):
Just a few doc edits and we're good.

Bilal2453:
I guess merging with squash is a good idea now.

(See on the right of the merge button, an arrow that shows a drop menu)
image

SinisterRectus:
I've always squashed recently.

SinisterRectus (approved):

PR #339, opened 2022-02-08
Use new `embeds` field for messages and allow multiple embeds.

Lyrth:
Use the new `embeds` field for sending messages since `embed` is now deprecated (see below), and allow sending multiple embeds in one message by specifying `embeds` in `TextChannel:send`. Specifying `embed` in the send method still works for compatibility, `embeds` will take precedence if it's a non-empty array.

( from: https://discord.com/developers/docs/change-log#june-10-2021 )
> Message routes now accept an embeds array in addition to the existing embed field. Bots can now send up to 10 embeds per message, to be consistent with webhook behavior. **The existing embed field is considered deprecated and will be removed in the next API version.**

### DIFF
--- a/libs/containers/abstract/TextChannel.lua
+++ b/libs/containers/abstract/TextChannel.lua
@@ -261,6 +261,14 @@ function TextChannel:send(content)
 			content = concat(mentions, ' ')
 		end
 
+		local embeds
+		if tbl.embed then
+			embeds = {tbl.embed}
+		end
+		if type(tbl.embeds) == 'table' and #tbl.embeds > 0 then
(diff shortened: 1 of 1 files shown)


RiskoZoSlovenska:
Perhaps it would be a good idea to support both the `embed` and `embeds` fields (at least from a user point of view), so that if both are present, the value of `embed` gets added to the `embeds` table. [That's how 3.0 does it](https://github.com/SinisterRectus/Discordia/blob/4ea9b86972d1cb82a80b1b0ee84b51149ef4aac6/libs/messaging.lua#L88-L96).

PR #347, opened 2022-05-26
Fix reply mention behaviour on message edit

object-Object:
This fixes a bug where if you sent a message with a reply with the mention disabled, editing the message caused the reply mention to become enabled.


SinisterRectus:
It's annoying that when you include `allowed_mentions` then the default is to not mention anything. This technically changes the expected behavior. But I think this is going to have to be the best fix for this.

SinisterRectus:
Thanks.

PR #354, opened 2022-07-22
doc-comment: fix some equivalent methods optional param

Bilal2453:
I've noticed many of the methods that are documented to be an "Equivalent" of another method imply the parameters are required, when they are not.   At first I thought that was intentional, but looking at the other methods that are "equivalent" they do document that (e.x. `Reaction:delete`), therefor and to be consistant the rest of those methods should follow that trend.

It also helps with the tools that relay on the doc-comment, such as [discordia-types](https://github.com/SovietKitsune/discordia-type) and [discordia-wiki-bot](https://github.com/Bilal2453/discordia-wiki-bot).

### DIFF
--- a/libs/containers/Member.lua
+++ b/libs/containers/Member.lua
@@ -435,7 +435,7 @@ end
 --[=[
 @m kick
 @t http
-@p reason string
+@op reason string
 @r boolean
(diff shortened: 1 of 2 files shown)


Bilal2453:
Note that this may conflict with my recent PRs.   I will rebase the PR if that happen.

PR #356, opened 2022-09-23
add component support

RemRBLX:


UrNightmaree:
Hmm 🤔

Bilal2453:
This also adds `embeds` fields effectively conflicting with PR #339, this should be removed from this PR.

I don't have a strong opinion about adding a `components` field here, although it doesn't make much sense to me when Discordia does not support Interactions, knowing that Message Components are useless most of the time without interactions, exception be it Link Buttons; Only way I can justify this is if you also add Interaction support.

Of course this would make more sense if you were using something like [discordia-interactions](https://github.com/Bilal2453/discordia-interactions), but if you are doing so you may as well use [discordia-components](https://github.com/Bilal2453/discordia-components).

Eventually, one may want to implement interactions, and this would be added nonetheless, so there is that going for it

PR #363, opened 2022-12-10
Update API version to v10

ben-forster:


truemedian:
API v10 is not a drop in replacement for v7, you will need to check the change logs between each version and figure out what needs to change so that Discordia sends correct requests.

PR #364, opened 2022-12-10
Update gateway version to v10 and voice API to v4

ben-forster:


truemedian:
The gateway has undergone massive changes since v6 (like intents), we cannot simply bump the version and expect anything to work.

PR #365, opened 2022-12-25
Slight UX enhancement for API Reference

ccuser44:
This makes reading the API reference a little easier by wrapping all of the function parameters/argments in the title wrapped in a code block


SinisterRectus:
Hi, can you provide an example output to show what this changes?

ccuser44:
Nevermind this wasnt as good as an idea I thought it was

PR #366, opened 2023-01-01
Happy new year!

Perthys:
:D


SinisterRectus:
Thank you, but I aim to update the license if the code is updated in that year or if a release is made in that year.

PR #367, opened 2023-01-08
2.x: Logger pretty newlines

RiskoZoSlovenska:
2.x backport of #320.

Additionally, unlike #320, this shouldn't introduce breaking changes to the return value of `Logger:log()`.

### DIFF
--- a/libs/utils/Logger.lua
+++ b/libs/utils/Logger.lua
@@ -14,7 +14,7 @@ file is desired. See the `logLevel` enumeration for acceptable log level values.
 local fs = require('fs')
 
 local date = os.date
-local format = string.format
+local format, gsub, rep = string.format, string.gsub, string.rep
 local stdout = _G.process.stdout.handle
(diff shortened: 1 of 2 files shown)


SinisterRectus:
I will review and possibly include this after 2.11.0.

libs/utils/Logger.lua
```diff
@@ -67,13 +68,22 @@ function Logger:log(level, msg, ...)
 	local tag = config[level]
 	if not tag then return end
 
+	local d = date(self._dateTime)
 	msg = format(msg, ...)
 
-	local d = date(self._dateTime)
+	local prettyfied = msg
```
SinisterRectus, line 74: No need to declare a new var here. Just re-use `msg`.

RiskoZoSlovenska: I wanted to preserve the original value of `msg` because it's the value that the method returns, and so it shouldn't include extra formatting.

PR #368, opened 2023-01-08
2.x: Client option to log full reason for HTTP errors

RiskoZoSlovenska:
2.x backport of #319.


SinisterRectus:
I will review and possibly include this after 2.11.0.

libs/client/API.lua
```diff
@@ -232,7 +233,12 @@ function API:commit(method, url, req, payload, retries)
 
 		end
 
-		client:error('%i - %s : %s %s', res.code, res.reason, method, url)
+		local errMsg = f('%i - %s : %s %s', res.code, res.reason, method, url)
+		if self._logFullErrors then
+			errMsg = errMsg .. "\n" .. msg
```
SinisterRectus, line 238: Shouldn't need to concat here. Take advantage of the logger's formatter.

libs/client/API.lua
```diff
@@ -112,9 +112,10 @@ end
 
 local API = require('class')('API')
 
-function API:__init(client)
+function API:__init(client, logFullErrors)
 	self._client = client
 	self._mutexes = setmetatable({}, mutexMeta)
+	self._logFullErrors = logFullErrors
```
SinisterRectus, line 118: Don't store this here. Just use the `options` table in the `commit` method.

libs/client/API.lua
RiskoZoSlovenska: I'm not sure what you mean. Something like
```lua
if logFullErrors then
    client:error(errMsg)
else
    client:error("%s\n%s", errMsg, msg)
end
```
?

PR #371, opened 2023-01-25
Fix for certain events not triggering (in v8?)

Lort533:
Fixes e.g. GUILD_MEMBER_ADD event not triggering at all (Discord won't send it in the first place).
Intent code generated thanks to https://discord-intents-calculator.vercel.app/. It would be best to verify if that intent code is actually the good one before pulling this request, because I only tested it on one event, and it worked.

Idea for fixing this issue came from here: https://stackoverflow.com/a/65640810

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -70,7 +70,7 @@ local defaultOptions = {
 	gatewayFile = 'gateway.json',
 	dateTime = '%F %T',
 	syncGuilds = false,
-	gatewayIntents = 3243773, -- all non-privileged intents
+	gatewayIntents = 3276799, -- all non-privileged intents
 }
 
 local function parseOptions(customOptions)


Bilal2453:
From https://github.com/SinisterRectus/Discordia/discussions/369:

> Privileged intents include: guildMembers, guildPresences, and messageContent. If you do not want to use privileged intents, no user code changes are required. [...] to use privileged intents in your bot, you must programmatically enable them in your bot code. To do this, you may use the enableIntents, enableAllIntents, or setIntents client methods after initilizing your Client instance, but before calling run.

So first of all, GUILD_MEMBER_ADD requires the Member privileged intent, hence Discordia does not enable it by default. It instead asks the user to do so with one of the provided methods (or in the client options).

For example you do:
```lua
local client = discordia.Client {
	gatewayIntents = 3276799,
}
```

> Since intents are now required and privileged intents are disabled by default, some side-effects may be seen. If you do not enable certain intents, your bot will simply never receive the events that are governed by those intents. 

Hence you don't receive the event at all, this also includes:

 - messageContent intent: If this is not enabled your bot will not receive the following fields and they will be empty strings: Message.content, Message.embeds, Message.attachment.
- guildPresences intent: If this is not enabled your bot will not receive presenceUpdate events.
- guildMembers intent: if this is not enabled your bot will not receive the following events: memberJoin, memberLeave, memberUpdate.


To summarize, this behavior is intentional and should not be changed, it was done to prevent bots that do not have the privileged intents enabled from breaking (the event in question is memberJoin); and instead the user needs to explicitly enable that in code after enabling it in the developer page.  I really advise to read https://github.com/SinisterRectus/Discordia/discussions/369 as you are transitioning into API v8 as it contains important clues.

Good luck

Lort533:
Okay, I haven't seen that announcement, I'm sorry for making this pull request.

Thank you.

PR #377, opened 2023-03-14
IP Discovery update

Lort533:
It seems to be working, prepared for change that will happen within a day from now.


Bilal2453:
[On this line](https://github.com/SinisterRectus/Discordia/pull/377/commits/4216e67a8ad5eca7ec0461d81e081695577ddaa5#diff-5cb13bbf92bb7d5fa615e986333af662ebc7a79d791b6b19532f39850d6fbc8dL170) you should change it to 8 bytes padding because you will receive an additional SSRC field which is 4 bytes.  Should become `xxxxxxxxz`.

Bilal2453:
LGTM

SinisterRectus:
I'll assume that this works. I hate voice code.

PR #378, opened 2023-03-23
Made "embeds" a valid field for Message:update

InZan17:
Allows for editing multiple embeds on a message.


SinisterRectus:
Embeds are fine, but it looks like you added allowed_mentions, too.

InZan123:
Sorry about that, I didn't mean to add that change to this pull request. I can remove it if you want.

PR #379, opened 2023-04-06
Change Discord from Freeware to Freemium

Hedwig7s:


AllMesi (approved):

SinisterRectus:
I'd rather leave the language as-is, or remove it entirely, instead of changing it. Yes, Discord has continuously augmented their Nitro offerings, but "freemium" is kind of a charged word and I really don't think this a place to make a point around it.

Issue #384, opened 2023-05-07
Permission management not functioning correctly


MeteorTheLizard:
The permission management for roles as well as categories and textChannels / voiceChannels does not function correctly. After a fair bit of research with the help of another discord user, it appears to be an issue with how the permission enums as well as everything further down the line is handled.

https://github.com/SinisterRectus/Discordia/blob/899a8d45bb8cdeed73656c786a7bd79179def79e/libs/enums.lua#L158
Should be `return 2ULL^n`

Changing it to this however causes a problem here as cdata is not recognized by the Resolver.
https://github.com/SinisterRectus/Discordia/blob/899a8d45bb8cdeed73656c786a7bd79179def79e/libs/client/Resolver.lua#L156

And when this function is expanded to support cdata, there is still the problem that cdata cannot be a table key.

Assuming the data type is preserved for permissions, changing the flag function in the enums.lua causes problems in other areas as the gatewayIntents starts to throw errors as well when changing it to `return 2ULL^n` - This I haven't looked into much further.

--

Stuff that I tried to do that lead to discovering the issue:

```lua
local tBotclass = require("../botinfo")

local enums = tBotclass.tDiscordia.enums.permission

local obj_Guild = message.guild

local obj_Role_Everyone

for _,v in pairs(obj_Guild.roles) do
	if v.name == "@everyone" then
		obj_Role_Everyone = v

		break
	end
end

local obj_Category = obj_Guild:createCategory("test-category-ignore")

local obj_Permissions = obj_Category:getPermissionOverwriteFor(obj_Role_Everyone)

obj_Permissions:denyAllPermissions() -- @everyone has no rights here!
```
Do note that I removed all error handling here to make this code easier to understand.

The code above results in this:

![image](https://user-images.githubusercontent.com/55365852/236704812-18ac160c-5a2d-4eca-b1a5-25bb88284296.png)

denyAllPermissions is clearly not working correctly.

When I attempt to deny specific permissions like so:

`obj_Permissions:denyPermissions(enums.readMessages,enums.changeNickname,enums.sendMessages,enums.embedLinks,enums.attachFiles,enums.addReactions,enums.useExternalEmojis,enums.readMessageHistory,enums.useSlashCommands)`

This happens:

![image2](https://user-images.githubusercontent.com/55365852/236704911-6feae0f1-45b4-4a35-a91e-52cf732849e5.png)

[...]

PR #386, opened 2023-05-11
Fix permissions on arm processors and prepare for permissions change

truemedian:
Works around a bug in luajit that seems to happen when passing a
number to `bit.bor` that is larger than 2^31-1

Additionally adds a few missing enumerations so that enableAll and
disableAll work as expected.

Fixes https://github.com/SinisterRectus/Discordia/issues/384

Also forces permissions to be sent as a string, in preparation for
when they'll no longer be allowed to be sent as a number.

### DIFF
--- a/libs/client/Resolver.lua
+++ b/libs/client/Resolver.lua
@@ -150,7 +150,7 @@ function Resolver.permissions(obj)
 	if isInstance(obj, classes.Permissions) then
 		return obj.value
 	end
-	return tonumber(obj)
+	return int(obj)
 end
 
 function Resolver.permission(obj)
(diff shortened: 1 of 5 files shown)


MeteorTheLizard:
I've been using this fix for a few weeks now and it has seen a lot of use. So far there haven't been any problems with it, though through testing I found out that when there is an enum missing in libs/enums.lua (for example when discord decides to add a new one on their end), then the permission management breaks again.

SinisterRectus:
It looks like this changes the enum type from number to LJ int. Not ideal, but is there no better way?

truemedian:
Closing as wont-fix. arm32 is relatively rare, and nobody has complained since the original report.

Additionally, this introduces a lot of other type-related issues and discord has stopped claiming that they'll stop at 64 bit integers so a better solution will need to be implemented in 3.0 anyway.

PR #388, opened 2023-05-12
PermissionOverwrite: use .type instead of ._type

Bilal2453:
Seems like an oversight. `PermissionOverwrite._type`'s type is an integer not a string, in the `PermissionOverwrite.type` getter this integer value is resolved into a string of either `member`/`role`.  For this comparison here seems like it was intended to be `.type`.


SinisterRectus:
Looks good, thank you.

PR #389, opened 2023-05-23
Add support for stickers

Ta-noshii:


SinisterRectus:
I guess that's better than nothing. What about uploading them? 

I will be able to properly review this after May 31.

Ta-noshii:
> I guess that's better than nothing. What about uploading them?
> 
> I will be able to properly review this after May 31.

Working on it!
I know this is pretty barebones, which is also why I left out the sending them part which I have in my version since its just the IDs in a table, ill get to work on making uploading stickers possible

Ta-noshii:
Everything here works fully so far, except creating stickers:
https://canary.discord.com/channels/749981577303556167/759401690943717438/1110720268747411557
I also added the new message flags to add support for silent messages, and custom allowed_mentions.

SinisterRectus:
This, and #378, both sneak "allowed_mentions" in. I'd rather keep that separated from the main goal of the PRs.

Ta-noshii:
Ill remove it when im home

PR #391, opened 2023-06-11
Fix typo in new username support update

Ta-noshii:


SinisterRectus:
Thank you, I'm sure this was a lot of work.

PR #400, opened 2023-10-11
No need to specify Bot prefix in token line

EtherealTheDev:
Makes it so you don't have to do "Bot token" but still allows for self-bots using bot param, 

bot=true, will do 'Bot ' + token
bot=false, will do just the token

Sorry if this description is incorrect or heard to read, its like 12 am

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -158,6 +158,8 @@ local function run(self, token, bot=true)
 		options.cacheAllMembers = false
 	end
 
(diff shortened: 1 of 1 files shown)


EtherealTheDev (commented):
Don't even know when I removed this

truemedian:
This isn't even valid Lua code. However, see the reasoning [the first time this was proposed](https://github.com/SinisterRectus/Discordia/pull/174#issuecomment-465250081).

Issue #404, opened 2024-01-29
Permission calculations in Member:getPermissions return incomplete results


BestMordaEver:
Permissions object returned by Member:getPermissions is mostly incomplete if a `channel` argument is passed. Example
```lua
print(guild.me:getPermissions())
-- Permissions: 140742071708752 (addReactions, connect, embedLinks, manageChannels, manageMessages, manageRoles, moveMembers, readMessages, requestToSpeak, sendMessages, speak)
print(guild.me:getPermissions(channel))
-- Permissions: 1049600 (connect, readMessages)
print(channel:getPermissionOverwriteFor(guild.me).allowedPermissions)
-- 1049600
print(channel:getPermissionOverwriteFor(guild.me).deniedPermissions)
-- 0
```
Issue was introduced 6493e89 - `permissions1:complement(permissions2)` replaced most instances of `band(permissions1, bnot(permissions2))`. In `complement`, the binary operation is `!a & b` instead of `a & !b`

PR #405, opened 2024-01-29
Fix for permissions calculation in Member:getPermissions

BestMordaEver:
fix for #404


SinisterRectus:
Thank you for your help!

PR #411, opened 2024-07-22
[new_changes]

Bonniaas:


Bonniaas:
Here is gonna be my First work as i try lua and maybe it gonna be help me to do with it! 🐰

cattokomo:
This probably need a context why your PR is closed.
1. Unnecesarry renaming.
2. Removing `ULL` suffix from integer, this is a valid syntax in LuaJIT and it's intended for working on 64-bit integers in Lua. (Lua 5.1, current implementation for LuaJIT doesn't have 64-bit integers until 5.3)

PR #412, opened 2024-07-24
Add support for Badges

NickIsADev:
This pull request creates a new Enum, `enum.userFlag`, creates a new Resolver, `Resolver.userFlag()`, a new User getter, `get.badges()`, and a new User method, `User:hasBadge(User-Flag-Resolvable)`. These all allow for Badges support.

- `get.badges()` returns an array of strings which represent the badge's name (which matches up with the Enum).
- `User:hasBadge()` returns a boolean indicating whether or not the user has the specified User-Flag-Resolvable.

Most of this was derived from Message Flags to maintain consistency.

Both the getter and the method have been tested.

One thing to note it that the `staff` badge does not seem to be able to be read due to its bit. I wasn't entirely sure what to do with the enum so I ended up keeping it. Another thing I wasnt so sure about was whether User:hasBadge() should return the flag value, or the badge name. It was agreed upon in the Discordia server that the badge name would be easier to handle.

The raw property of a User that provides the functionality of Badges is `User._public_flags`. The documentation can be found [here](https://discord.com/developers/docs/resources/user#user-object-user-flags).


SinisterRectus:
This looks great except I would prefer to use the API term "flag" everywhere instead of using the product term "badge". So, `user:hasFlag` instead of `user:hasBadge` and `user.flags` instead of `user.badges`.

SinisterRectus:
I'm not sure I like the `user.badges` or `user.flags` property, but it's probably the best we can do. For permissions, we have a dedicated class so that `member:getPermissions():toArray()` gives the list of names, but I'm not about to recommend a `UserFlag` class. This is expected to handled more generically in 3.x with a `Bitfield` class that can represent any type of bitfield/flag/enum beyond just permissions.

libs/containers/User.lua
```diff
@@ -202,4 +219,18 @@ function get.mutualGuilds(self)
 	return self._mutual_guilds
 end
 
+--[=[@p badges Array An array of all badges the user has, represented by the badge's name.]=]
```
Bilal2453, line 222: This should either say `table` or it should return an `ArrayIterable` instead of a table (I recommend this), `Array` on its own is not a valid type.

libs/containers/User.lua
```diff
@@ -57,6 +62,18 @@ function User:getDefaultAvatarURL(size)
...
+--[=[
+@m hasBadge
+@t mem
+@p badge User-Flag-Resolvable
+@r boolean
+@d Indicates whether the user has the badge given.
+]=]
+function User:hasBadge(badge)
+	badge = Resolver.userFlag(badge)
+	return band(self._public_flags or 0, badge) == badge
```
Bilal2453, line 74: Note that this method works differently compared to something like `Message:hasFlag()`, which checks for the presence of any of the provided flags. Might want to change it to how Message does it, or vise versa.

NickIsADev:
I will redo this at some point.

PR #414, opened 2024-08-28
fix for deleted objects not getting updated with new data upon recreation

BestMordaEver:
Is there a specific reason `Cache:_insert` doesn't apply `:_load` to deleted objects? If no, then this PR should fix that.


SinisterRectus:
I don't think I ever omitted this with a good reason.

PR #416, opened 2024-09-20
Various voice updates in preparation for november deprecation

truemedian:
The old `xsalsa20_poly1305` encryption mode is being removed finally. This leaves 2 encryption modes remaining, the required `aead_xchacha20_poly1305_rtpsize` and recommended `aead_aes256_gcm_rtpsize`. AEAD AES256-GCM is only recommended because libsodium only provides support for it when the hardware on the system supports AES instructions. If the hardware does not provide this support, Discordia will automatically fall back to only requesting the AEAD XChaCha20 Poly1305 encryption.

The update to these new encryption modes required an update of our vendored libsodium binaries. Both of these encryption modes have existed in libsodium for at least 8 years and most package managers should provide a libsodium with these functions.

In hope of preventing mistakes when handling encrypted RTP packets, encoding and decoding of the RTP layer has been abstracted into functions in VoiceConnection, this primarily decreases the chance of extensions like discordia-voice handling a change in encryption modes incorrectly, and making mistakes much less likely.

In the meantime, the voice gateway has also been updated to V8, this primarily introduces buffered resume, which re-sends any lost websocket messages on resume.


SinisterRectus:
I'm testing this on Windows 11 using the sodium and opus binaries that we provide. 

> sodium.lua:91: cannot resolve symbol 'crypto_aead_xchacha20poly1305_ietf_messagebytes_max': The specified procedure could not be found.

Do we need an update?

SinisterRectus:
Disregard that, I figured it out. We do need an update.

SinisterRectus:
https://download.libsodium.org/libsodium/releases/

libsodium-1.0.20-msvc.zip works on my machine

truemedian:
The updated binaries in 48d4d4c4b5fdff6c57456917d8a5b83fb7a656a4 were taken directly from libsodium-1.0.20-stable-msvc.zip

SinisterRectus:
Oh you did update them. My bad.

SinisterRectus:
Thank you for this work.

PR #420, opened 2024-10-16
Date: support discord's timestamp mention format

Bilal2453:
Added the follow new methods to Date class:

- `Date.parseTimestampFormat(str)`, returns `number, string`.
- `Date.fromTimestampFormat(str)`, returns `Date`.
- `Date:toTimestampFormat(style)`, accepts `enums.timestampStyle`, returns string.

Might want to choose a better name than `TimestampFormat`, but that is what [Discord calls it internally](https://discord.com/developers/docs/reference#message-formatting) it seems.

### DIFF
--- a/libs/enums.lua
+++ b/libs/enums.lua
@@ -324,4 +324,14 @@ enums.logLevel = enum {
 	debug   = 4,
 }
 
+enums.timestampStyle = enum {
+	shortTime      = 't',
+	longTime       = 'T',
+	shortDate      = 'd',
+	longDate       = 'D',
+	shortDateTime  = 'f',
+	longDateTime   = 'F',
+	relativeTime   = 'R',
+}
+
 return enums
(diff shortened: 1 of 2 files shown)


SinisterRectus:
3.0 calls it a mention

- https://github.com/SinisterRectus/Discordia/blob/dev/libs/messaging.lua#L28
- https://github.com/SinisterRectus/Discordia/blob/dev/libs/utils/Date.lua#L263

SinisterRectus:
I prefer a name like `toTimestamp` or `toDiscordTimestamp`. The "Format" in `toTimestampFormat` is not consistent with other method names like `toUTC` which is not named `toUTCFormat`.

Bilal2453:
I agree. I do like `toTimestamp` but it may be confusing to the user considering there are multiple references to a "timestamp". `toMention` and `toDiscordTimestamp` are fine, I will have it as `toDiscordTimestamp`.

Bilal2453:
Do you want something like `Message.mentionedTimestamps`? I think it is unnecessary.

PR #423, opened 2024-10-18
feat: Add crosspost and news channel supports

Bilal2453:
Adds the ability to follow Announcement/News channels, as well as the ability to crosspost messages.
This should be forward compatible with the dev branch.

### New Methods

- `Message:crosspost()` - cross-posts the message publishing it into following channels

- `GuildTextChannel:follow(targetId)` - start following the news channel and publishing announcements into `targetId`.

Note: to access the original message of a cross-posted message `referencedMessage` can be used. Currently the docs for it are wrong, `referencedMessage` is more than just replies.

### DIFF
--- a/libs/containers/GuildTextChannel.lua
+++ b/libs/containers/GuildTextChannel.lua
@@ -118,6 +118,26 @@ function GuildTextChannel:enableNSFW()
 end
 
 --[=[
+@m follow
+@t http
+@p targetId Channel-ID-Resolvable
+@r string
+@d Follow this News channel and publish announcements to `targetId`.
+Returns a 403 HTTP error if `GuildTextChannel.isNews` is false.
+]=]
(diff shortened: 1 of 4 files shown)


libs/containers/GuildTextChannel.lua
```diff
@@ -117,6 +117,25 @@ function GuildTextChannel:enableNSFW()
...
 end
 
+--[=[
+@m follow
+@t http
+@r string
+@d Follow this News channel and publish announcements to `targetId`.
+Returns a 403 HTTP error if `GuildTextChannel.isNews` is false.
+]=]
+function GuildTextChannel:follow(targetId)
```
SinisterRectus, line 127: Need to document `targetId` parameter.

Bilal2453: oops, good catch

PR #424, opened 2024-10-19
chore: Sticker: fix docs and general cleanup

Bilal2453:
Just a bit of documentation fix.

I noticed this
```lua
--[=[@p stickers ArrayIterable An iterable array of all stickers that are sent in this message.]=]
function get.stickers(self)
	if not self._stickers then
		local client = self.client
		self._stickers = ArrayIterable(self._sticker_items, function(sticker)
			if sticker.format_type == 1 then
				local guild = client._sticker_map[sticker.id]
				return guild and guild._stickers:get(sticker.id) or nil
			else
				-- return client:getSticker(sticker.id) ??
			end
		end)
	end
	return self._stickers
end
```
There should be no reason to only load stickers of type `1`, but I will look into it properly later.

### DIFF
--- a/libs/containers/Sticker.lua
+++ b/libs/containers/Sticker.lua
@@ -13,16 +13,10 @@ local Sticker, get = require('class')('Sticker', Snowflake)
 function Sticker:__init(data, parent)
 	Snowflake.__init(self, data, parent)
 	self.client._sticker_map[self._id] = parent
-	return self:_loadMore(data)
 end
 
 function Sticker:_load(data)
 	Snowflake._load(self, data)
-	return self:_loadMore(data)
-end
-
-function Sticker:_loadMore(data)
-	
 end
 
 function Sticker:_modify(payload)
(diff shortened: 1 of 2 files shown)


libs/containers/Sticker.lua
```diff
@@ -13,16 +13,12 @@ local Sticker, get = require('class')('Sticker', Snowflake)
 function Sticker:__init(data, parent)
 	Snowflake.__init(self, data, parent)
 	self.client._sticker_map[self._id] = parent
-	return self:_loadMore(data)
+	return nil
```
SinisterRectus, line 16: Returning `nil` from initializers is not necessary. It is used for "loading more" as a tail call optimization.

libs/containers/Sticker.lua
```diff
@@ -13,16 +13,12 @@ local Sticker, get = require('class')('Sticker', Snowflake)
...
 end
 
 function Sticker:_load(data)
 	Snowflake._load(self, data)
-	return self:_loadMore(data)
-end
-
-function Sticker:_loadMore(data)
-	
+	return nil
```
SinisterRectus, line 21: Returning `nil` from initializers is not necessary. It is used for "loading more" as a tail call optimization.

PR #428, opened 2024-11-05
docs: change Intesnts-Resolvables to Gateway-Intent-Resolvable

Bilal2453:
I don't know if this is a change that makes sense for you, so feel free to close it if not. But here is the analogy behind it: from what I can see resolvables use singular nouns for when they represent *one* thing of the type, for example `Permission-Resolvable` vs `Permissions-Resolvable`, in this case, the resolvable represents a single value only (many of those singular values are accepted as varargs though), so the singular noun should be used instead.

I am also adding `gateway` to make it clear that this also corresponds with the `gatewayIntent` enumeration.

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -329,7 +329,7 @@ end
 --[=[
 @m setIntents
 @t mem
-@p intents Intents-Resolvable
+@p intents Gateway-Intent-Resolvable
 @r nothing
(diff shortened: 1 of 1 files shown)


SinisterRectus:
The enum names are intentionally singular so that e.g. `gatewayIntent.messageContent` is read as: "the message content gateway intent". Permissions and Intents is plural because they can represent multiple values in one.

PR #430, opened 2024-11-05
docs: fix intents-related docs

Bilal2453:
1. `setIntents` parameter type changed to `number` since it is never resolved in a meaningful way.
2. `setIntents` paramter changed to optional, as it will default to `0`.
3. changed `Intents-Resolvables` to `Intents-Resolvable` because that is what Discordia uses everywhere.

### DIFF
--- a/libs/client/Client.lua
+++ b/libs/client/Client.lua
@@ -330,10 +330,10 @@ end
 --[=[
 @m setIntents
 @t mem
-@p intents Intents-Resolvable
+@op intents number
 @r nothing
 @d Sets the gateway intents that this client will use. The new value will not be
-used internally until the client (re-)identifies.
+used internally until the client (re-)identifies. Defaults to `0`.
 ]=]
 function Client:setIntents(intents)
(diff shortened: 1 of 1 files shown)


SinisterRectus:
I already said I didn't want to change this. 1 and 3 are intentional. 2 is fine.

Bilal2453:
Sorry, seems like it was a miscommunication! I thought you *wanted* this.

Then nothing needs to be changed here, but there is a bit of a consistency problem to describe here:

Currently in the Input resolution wiki page we list all resolvers, the way those Input resolution entries are named is by *removing* the `-Resolvable` suffix, leaving the resolver name, so for example `Permission-Resolvable` becomes `Permission`, `Permissions-Resolvable` becomes `Permissions` etc.  If we use the same pattern here, we end up with two entries with the same name, `Intents` and `Intents`, because the difference is in the suffix which we currently remove.

If we use the same pattern we use right now, you will end up with two entries having the same name like this:
image

PR #431, opened 2024-11-07
fix: enum and docs typos

Bilal2453:
Various typos I've noticed while working on a type def gen.

The enum typo has sadly made it into the latest release, but hopefully isn't too late to fix it now.


SinisterRectus:
Technically can't change the current one, but the new one can be added.

Bilal2453:
readded the enum with a deprecated comment

PR #434, opened 2025-02-25
use ucrt compiled libsodium to avoid dependency issues

truemedian:
Some time between when the sodium binaries were originally uploaded years ago and the recent update to libsodium 1.20 the prebuilt libsodium binaries introduced non-ideal dependencies. Sodium provides windows binaries using both MSVC and MinGW, the MSVC-built binaries require `vcruntime140` which requires users to have downloaded and installed the MSVC redistributable; the MinGW-built binaries require `libgcc_s_seh-1`, which is generally only present alongside a MinGW installation.

Neither of these dependencies were originally required and it would be preferable to keep it this way. Getting rid of these dependencies is possible by forcing the MSVC build to use UCRT and therefore not depend on the MSVC runtime, but this means these binaries are *not* prebuilt binary releases.

These binaries were compiled on my machine, by me, around 2 months ago; if you operate on a system that requires strict scrutiny of dependencies then you likely should be downloading these binaries from the source anyway.


SinisterRectus:
Are these the MSVC binaries depending on UCRT?

Bilal2453:
The binaries provided in this PR are the UCRT ones, compiled with MSVC. Which works without needing to install anything else as (unlike the MSVC/GCC runtime) of Windows 10.

PR #439, opened 2026-02-22
fix(gateway): use resume_gateway_url when reconnecting

NickIsADev:
This pull requests implements the missing handling of `resume_gateway_url`, which was added to the Gateway in August of 2022 (https://docs.discord.com/developers/change-log#session-specific-gateway-resume-urls). The system was pushed into full use later in September of the same year.

This system does not exist for the Voice Gateway at the moment, hence why no changes were made there.

Changes:
- Store `Shard._resume_gateway_url` when receiving it via the `READY` event
- Use `Shard._resume_gateway_url` to reconnect if available, otherwise use original gateway URL

Documentation Reference: https://docs.discord.com/developers/events/gateway#preparing-to-resume

### DIFF
--- a/libs/client/EventHandler.lua
+++ b/libs/client/EventHandler.lua
@@ -52,6 +52,7 @@ function EventHandler.READY(d, client, shard)
 	shard:emit('READY')
 
 	shard._session_id = d.session_id
+	shard._resume_gateway_url = d.resume_gateway_url
 	client._user = client._users:_insert(d.user)
 
 	local guilds = client._guilds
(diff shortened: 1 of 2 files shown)


NickIsADev:
I've also noticed that Discordia does not properly produce jitter on the initial heartbeat, however I'm unsure of how to approach this issue while keeping the code clean and readable. I've thought of just using `sleep(heartbeat_interval * math.random)` or `setTimeout`, however that could lead to even more issues with many heartbeat timers running if multiple reconnects happen quickly (for whatever reason).

Documentation Reference: https://docs.discord.com/developers/events/gateway#heartbeat-interval

